PR #1, opened 2019-10-07
Create front end project based on youth-membership-ui

badrange:


.eslintrc.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": ["react-app", "plugin:@typescript-eslint/recommended"],
```
Nizar-Rahme: Don't know about you, but I'm a fan of Prettier and usually add it as a plugin in `eslint` config.

midori0507: We will add them later after this.

PR #8, opened 2019-10-10
KK-34 | Styles

midori0507:
- Add alias for easier sass var import.
- Add normalizer css, inject Helsinki fonts.
- Extract basic style from opencity.design.

Extra: Join CI jobs to reduce machine booting time -> faster CI


.env
```diff
@@ -0,0 +1 @@
+SASS_PATH=./src/assets/styles
```
Nizar-Rahme: `.env` file shouldn't be pushed to GitHub

src/domain/app/header/Header.module.scss
```diff
@@ -0,0 +1,24 @@
...
+.appName {
+  margin-left: 30px;
+}
+
+.logo {
+  @include helsinki-logo($hel-black);
+  background-position: center;
+  width: $logo-width;
+  height: inherit;
+}
```
Nizar-Rahme: Wouldn't it be better to stick to the naming convention with other projects, which is starting `scss` files with small letters?

.env
jimppa: Actually it should if following CRA: https://create-react-app.dev/docs/adding-custom-environment-variables#adding-development-environment-variables-in-env

midori0507: They should, according to create-react-app docs :D https://create-react-app.dev/docs/adding-custom-environment-variables#docsNav

Nizar-Rahme: Fair enough

src/domain/app/header/Header.module.scss
midori0507: Done

PR #9, opened 2019-10-10
KK-36 | Add Sentry to project

badrange:
Initial configuration for sentry. 

We only send sentry reports if the environment is set, for local development you won't set it.


.env
```diff
@@ -0,0 +1,3 @@
+REACT_APP_APPLICATION_NAME=$npm_package_name
+REACT_APP_VERSION=$npm_package_version
```
midori0507: We can have first 2 variable in `.env.local`. Which will populate for all env, except `.env.test` will have to manually copy.

.env.local.example
```diff
@@ -0,0 +1,3 @@
+# If you for some reason want Sentry reports to be sent from
+# your local computer, you can set this variable:
+#REACT_APP_SENTRY_ENVIRONMENT=local
```
midori0507: This file is not needed

.env.test
```diff
@@ -0,0 +1 @@
+REACT_APP_SENTRY_ENVIRONMENT=test
```
midori0507: This file is not needed

PR #10, opened 2019-10-15
KK-37 | CSS Grid Responsive Layout

midori0507:
- Create responsive layout with vanilla css-grid
- Create exportable formatMessage


src/assets/styles/variables.scss
```diff
@@ -0,0 +1,5 @@
+$headerHeight: 4em;
+$footerHeight: 16em;
+
+$base-padding: 1em;
```
Nizar-Rahme: I prefer relying on `rem` instead of `em`

PR #11, opened 2019-10-16
KK-38 | Create Home layout

midori0507:
- Add Button and Input reusable components.
- Add formik (for later usage).
- Add classnames.

- Add couple snapshot tests


src/common/components/button/Button.tsx
```diff
@@ -0,0 +1,19 @@
...
+import joinClassNames from 'classnames';
+
+import styles from './button.module.scss';
+
+const Button: FunctionComponent<{
+  className?: string;
+  children: string | ReactElement;
+}> = ({ children, className, ...rest }) => {
+  return (
+    <div className={joinClassNames(styles.buttonWrapper, className)}>
```
Nizar-Rahme: Why don't you apply these class names directly on the `button` element?

src/common/components/button/button.module.scss
```diff
@@ -0,0 +1,14 @@
+@import "colors";
+@import 'variables';
+
+.buttonWrapper {
+    display: flex;
```
Nizar-Rahme: Not sure why flex is needed here

midori0507: i dont think grid is needed all the time

src/common/components/button/Button.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { FunctionComponent, ReactElement } from 'react';
+import joinClassNames from 'classnames';
+
+import styles from './button.module.scss';
+
+const Button: FunctionComponent<{
+  className?: string;
+  children: string | ReactElement;
```
Nizar-Rahme: How about?
```suggestion
  children: React.ReactNode;;
```

midori0507: i want vanilla `<button>foo</button>` work here too, not stricted to only jsx

src/common/components/button/button.module.scss
Nizar-Rahme: Yeah, I think neither `flex` nor `grid` is needed here.

src/common/components/button/Button.tsx
midori0507: I think it's better to have wrapper, rather than modifying straight on the html5 `button`.
Sometime they need some padding / margin / customization / outside label

src/common/components/button/Button.tsx
Nizar-Rahme: Yes, as far as I know, `ReactNode` includes strings, nulls, JSX elements, etc.

midori0507: ah, i'll remove it then

src/common/components/button/button.module.scss
midori0507: agree, i just want to make sure it can take all from parents

src/common/components/button/Button.tsx
midori0507: `Type 'string' is not assignable to type 'ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any)`

midori0507: fixed

PR #15, opened 2019-10-22
KK-40 | Create a script that fetches translations from Google Spreadsheets

badrange:
* I made the script configurable through `.env` so that it is easy to reuse
* There is a sanity check to ensure that getting the CSV in the right language actually works 

To test, run 

`yarn update-translations`


badrange:
Example of how the sheet looks like
image

badrange:
image

Example of how the sheet looks like

PR #19, opened 2019-10-29
KK-55 | Private route

midori0507:
- Create PrivateRoute, check for auth -> redirect.
- Create loading spinner to halt the app until user data was fetched.

- Add additional check for string in validation utils -> uppercase letter will not be matter


src/domain/auth/state/authSelectors.tsx
```diff
@@ -0,0 +1,14 @@
+import { createSelector } from 'reselect';
+
+import { StoreState } from '../../app/types/stateTypes';
+
+export const idTokenSelector = (state: StoreState) =>
+  state.authentication.user && !!state.authentication.user.id_token;
```
jimppa: You probably want to remove `!!` from `!!state.authentication.user.id_token;` here

midori0507: true, i tested and forgot to remove, i merge too fast, thank for pointing out!

jimppa: I'll try to be faster next time!

PR #21, opened 2019-10-30
KK-29 | Add graphql to the project

badrange:
- add script that fetches typings from schema (    `"graphql-types": "apollo client:codegen --target=typescript --globalTypesFile='src/__generated__/globalTypes.ts'"`
)


badrange:
- add script that fetches typings from schema (    "graphql-types": "apollo client:codegen --target=typescript --globalTypesFile='src/__generated__/globalTypes.ts'"
)
-

apollo.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  client: {
+    service: {
+      name: 'kukkuu',
+      url: 'http://localhost:8001/graphql',
```
midori0507: config.js ? -> .ts instead?

package.json
```diff
@@ -73,6 +79,7 @@
       "src/**/*.{ts,tsx}",
       "!<rootDir>/node_modules/",
       "!src/index.tsx",
+      "!src/domain/graphql/generated",
```
midori0507: Make `src/domain/api` instead of `graphql` name

src/domain/graphql/generated/globalTypes.ts
```diff
@@ -0,0 +1,11 @@
+/* tslint:disable */
+/* eslint-disable */
+// This file was automatically generated and should not be edited.
+
+//==============================================================
```
midori0507: if folder name can change from `generated` to `generatedTypes` would be nice

src/domain/auth/userManager.ts
```diff
@@ -8,6 +8,7 @@ const location = `${window.location.protocol}//${window.location.hostname}${
 /* eslint-disable @typescript-eslint/camelcase */
 const settings: UserManagerSettings = {
   authority: process.env.REACT_APP_OIDC_AUTHORITY,
+  automaticSilentRenew: true,
```
midori0507: nope

src/domain/home/form/HomePreliminaryForm.tsx
```diff
@@ -32,6 +33,7 @@ interface HomeFormValues {
 
 interface Props {
   setFormValues: (values: RegistrationFormValues) => void;
+  submittedFormValues: RegistrationFormValues;
```
midori0507: Why you connect the data but you dont use it, should be removed

apollo.config.js
badrange: Required adding `export {}` to the file to address ts complaint: 

> Cannot compile namespaces when the '--isolatedModules' flag is provided

apollo.config.ts
```diff
@@ -0,0 +1,9 @@
+module.exports = {
```
midori0507: use ES6 export
`
export default {content}
`

package.json
midori0507: This has not been updated with new path

midori0507:
@badrange You might want to include script in README doc? not gonna hurt anybody

package.json
badrange: Oo! Good spot!

apollo.config.ts
badrange: That doesn't work with the apollo command. Changing back to js file

PR #22, opened 2019-11-04
KK-28 | Registration form scaffolding.

midori0507:
- Upgrade formik to latest.
- Add scaffolding skeleton for registration form.
- Rename configuration files like reducer / typedef into `ts` extension instead of `tsx`


src/common/components/form/fields/input/InputField.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
-import { FieldProps } from 'formik';
+import { FieldProps, getIn } from 'formik';
```
jimppa: I see that `getIn` is exported and used in formiks documentation, but it is not documented and I don't think it's a good idea to use undocumented APIs from libraries. Would it be possible to use `lodash/get` or copy this function locally?

src/domain/home/form/HomePreliminaryForm.tsx
```diff
@@ -9,133 +9,153 @@ import Button from '../../../common/components/button/Button';
...
+    birthday: {
+      day: string | number;
+      month: string | number;
+      year: string | number;
+    };
+    homeCity: string;
+  };
   verifyInformation: boolean;
-  childBirthday?: string;
+  childBirthDay?: string;
```
jimppa: Birthday is one word, no need to change.

midori0507: good one!

src/common/components/form/fields/input/InputField.tsx
midori0507: Its a library thing, `getIn` is some kind of `formik` ulti, errors from field sometime can be array (fieldarray), which kinda complicated to handler with lodash with 1 fallback value only

src/domain/app/BrowserApp.tsx
```diff
@@ -21,15 +22,17 @@ if (process.env.NODE_ENV !== 'production') {
...
-  <Switch>
-    <Route exact path="/callback" component={OidcCallback} />
-    <Redirect exact path="/" to="/fi/home" />
-    <Route path={`/${localeParam}/*`} component={App} />
-    <Route exact path={`/${localeParam}/callback`} component={OidcCallback} />
-    <Route
-      render={props => <Redirect to={`/fi${props.location.pathname}`} />}
-    />
-  </Switch>
+  <PageLayout>
```
jimppa: You should be very sure that the layout is the same for all pages before automatically applying the layout by default on all routes, usually it's better to use the layout on every route "manually".

midori0507: Yeah i've been consider about that but for now i feel like every page should have header and footer so that's why (even a full page spinner). Spamming the layout was kinda more customizable but not needed here. 

I will remove the wrap and do manually in case we need one in future.

src/domain/registration/state/RegistrationSelectors.tsx
```diff
@@ -0,0 +1,4 @@
+import { StoreState } from '../../app/types/stateTypes';
```
jimppa: Wouldn't `.ts` be the correct file-extension for this file?

src/domain/registration/types/RegistrationTypes.tsx
```diff
@@ -1,7 +1,18 @@
 export interface RegistrationFormValues {
-  childBirthday: string;
-  childHomeCity: string;
+  child: {
```
jimppa: I think this file should be `.ts`-extension as well

midori0507: Wont hurt if with/without `x`
Sometime util return html -> x is wider choice

midori0507: but yea the typing should be ts

src/domain/registration/form/RegistrationForm.tsx
```diff
@@ -1,5 +1,205 @@
...
+                  name="guardian.relationship"
+                  label={formatMessage(
+                    'registration.form.guardian.relationship.input.label'
+                  )}
+                  onChange={handleChange}
+                  value={values.guardian.relationship}
+                  component={SelectField}
+                  id="registration.form.guardian.relationship.select"
+                  options={[
+                    { label: 'Parents', value: 'parents' },
```
jimppa: Any reason why label is not translated?

midori0507: its just a mock, will remove later when we go specific, but i can make a translation to be more legit code

src/domain/registration/state/RegistrationSelectors.tsx
midori0507: Agree

src/domain/registration/form/RegistrationForm.tsx
jimppa: Ok, then it's probably unnecessary to translate it now

src/domain/registration/state/RegistrationSelectors.tsx
midori0507: Would reducer should be `.ts` too?

jimppa: I think it should only be tsx/jsx when the file has some jsx in it, so usually just components (or their tests)

PR #25, opened 2019-11-05
KK-77 | Add silent_renew token when expired, auto-fetch apiToken when accesstoken exist

midori0507:
- Add fetch api token thunk and actions 
- Fetch apiToken when access token is available.

- Enable silently renew token ->trigger loadUser change -> renew apiToken.


.eslintrc.json
```diff
@@ -30,7 +30,6 @@
     "no-console": "warn",
     "no-plusplus": "error",
     "no-undef": "warn",
-    "object-curly-spacing": ["warn", "always"],
-    "sort-keys": ["warn", "asc", { "caseSensitive": true, "natural": true, "minKeys": 2 }]
```
badrange: 😿

src/domain/api/constants/ApiConstants.ts
```diff
@@ -0,0 +1,2 @@
+export const TUNNISTAMO_ENDPOINT = 'http://localhost:8000';
```
badrange: Shouldn't this be in an .env-file? I guess this is still wip :-)

badrange: The readme uses `tunnistamo-local` as hostname

src/domain/auth/userManager.ts
```diff
@@ -8,9 +8,11 @@ const location = `${window.location.protocol}//${window.location.hostname}${
 /* eslint-disable @typescript-eslint/camelcase */
 const settings: UserManagerSettings = {
   authority: process.env.REACT_APP_OIDC_AUTHORITY,
+  automaticSilentRenew: true,
   client_id: process.env.REACT_APP_OIDC_CLIENT_ID,
   redirect_uri: `${location}/callback`,
   response_type: 'id_token token',
+  silent_redirect_uri: `${location}/silent_renew`,
```
badrange: We need to find a way to easily add this to tunnistamo's callback url list/add it to the readme

.eslintrc.json
midori0507: It doesnt have autofix, really annoying.. I've read some debate of removing it by devs also

src/domain/api/constants/ApiConstants.ts
midori0507: Yeah i will use env value

.eslintrc.json
badrange: Yes, I will not object to removing this.

src/domain/api/client.ts
```diff
@@ -13,7 +13,7 @@ export default new ApolloClient({
       });
     } catch (e) {
       // User not authenticated
-      console.error(e);
+      // TODO: add error-handler
```
badrange: Is this swallowing the error? I think we should leave the console.error otherwise we might spend hours scratching our heads

PR #32, opened 2019-11-12
KK-78 | Add login logout button as well as redirection

midori0507:
- Add login / logout
- Add Dropdown component

- Logout redirection somehow doesnt send user back to expect URL, delegate this problem to Tunnistamo OIDC


src/domain/auth/authenticate.ts
```diff
@@ -10,9 +10,13 @@ import {
 import { TUNNISTAMO_API_TOKEN_ENDPOINT } from '../api/constants/ApiConstants';
 import { BackendTokenResponse } from './types/BackendAuthenticationTypes';
 
-export default function(path: string): void {
+export const loginToTunnistamo = (path?: string) => {
```
badrange: Naming is not consistent with `logoutTunnistamo` 

What if we call them: `loginToTunnistamo` and `logoutFromTunnistamo` or just `loginTunnistamo` and `logoutTunnistamo`

or (my favourite) `loginViaTunnistamo` & `logoutViaTunnistamo`?

midori0507: loginTo and loginFrom are more cute i would say

PR #33, opened 2019-11-13
KK-29 | Refactor birthday into birthdate

badrange:


badrange:
I have changed the keys in the translation spreadsheet

PR #37, opened 2019-11-18
Refactor required input field

midori0507:


midori0507:
> Did Anniina say that we wanted the asterisk?

As i just asked, yes

src/domain/home/form/HomePreliminaryForm.tsx
```diff
@@ -83,7 +83,8 @@ const HomePreliminaryForm: FunctionComponent<Props> = ({
         initialValues={convertFormValues(stateFormValues)}
         onSubmit={handleSubmit}
         validate={validate}
-        isInitialValid={false}
+        initialErrors={{ childBirthdate: validateBirthdate('') }}
+        // To make sure form will not be valid to submit at first load
```
badrange: Isn't it more normal to have the comment above the line it is about?

midori0507: Yea i will put it above

PR #43, opened 2019-11-21
KK-117 | Multiple children in graphql mutation

badrange:
This PR seems to work with the backend and only one child. 

Now that the backend has a more "complete" schema it might be possible to create the mutations in a more fancy apollo + graphql + typescript 2019 kinda way.


src/domain/registration/mutations/submitChildrenAndGuardianMutation.ts
```diff
@@ -1,32 +1,27 @@
...
           email: $email
         }
       }
     ) {
-      child {
+      children {
         birthdate
         firstName
         lastName
+        firstName
```
midori0507: duplicated firstName

PR #50, opened 2019-11-29
KK-123 | Add key to child list

badrange:
This removes a warning about missing keys in browser developer console
We need those keys for crud operations too


midori0507:
@badrange: This is quite not needed, you can use element index for it

badrange:
@midori0507 I believe we will need it for the child crud?

PR #54, opened 2019-12-02
KK-125 | Add a proper GuardianInput for better typing

badrange:


src/domain/registration/form/RegistrationForm.tsx
```diff
@@ -61,18 +60,28 @@ const RegistrationForm: FunctionComponent<Props> = ({
               setFormValues(values);
 
               // FIXME: Ensure that relationship is submitted to backend
-              const backendSupportChildren = values.children.map(child =>
-                omit(child, ['postalCode', 'homeCity', 'relationship'])
-              );
+              const backendSupportChildren = values.children.map(child => {
+                const c = omit(child, [
```
midori0507: you dont need variable c, plus extra unnecessary texts
Remove all the extra, the left side is already fine, no need to change.

src/domain/registration/form/RegistrationForm.tsx
```diff
@@ -61,18 +60,28 @@ const RegistrationForm: FunctionComponent<Props> = ({
...
+                  'postalCode',
+                  'homeCity',
+                  'relationship',
+                ]);
+                // Here is the meat of relationshiop:
```
midori0507: bad comments

midori0507: We should actually change the relationships into array instead of doing duct-tape

src/domain/registration/form/RegistrationForm.tsx
badrange: At the moment we are not able to submit relationship info to the backend, I don't think the left side is fine. But sure, we can find a different way of doing that.

badrange: I reverted this change. We have to fix this very soon, but in a different PR

PR #60, opened 2019-12-04
KK-135 | Add a custom translated title for each page

badrange:
BONUS: The Not eligible page does not link to the front page any more.


src/domain/profile/Profile.tsx
```diff
@@ -28,19 +29,26 @@ const Profile: FunctionComponent = () => {
...
-    <Container className={styles.grayBackground}>
-      <div className={styles.profileWrapper}>
-        <h1>{t('profile.heading')}</h1>
-        <div className={styles.childInfo}>
-          {profile &&
-            profile.children.map(child => (
-              <p key={child.id}>{child.firstName}</p>
-            ))}
-        </div>
+    <HelmetProvider>
```
midori0507: Wrap the Layout with the Provider, not each of page

badrange: That requires me to put the <Helmet> inside the <Container> on the Home page for example - are you ok with that?

PR #62, opened 2019-12-05
Registration form preferred language should default to UI language

badrange:


src/domain/registration/form/RegistrationForm.tsx
```diff
@@ -216,6 +217,7 @@ const RegistrationForm: FunctionComponent<Props> = ({
                   </div>
 
                   <EnhancedInputField
+                    value={values.preferLanguage || currentLocale}
```
midori0507: you should feed this value to initialValues, not here

midori0507: (hint) do that in selector

PR #63, opened 2019-12-05
KK-146 | Add error handling using react-toastify

badrange:
We should probably check with  Anniina how we should handle the different errors. This is a good step forward, though. 

For example: When profile fails to load due to network being down we might not necessarily want to send people to the front page. For now I created a broken looking error div. 

We also want to avoid sending errors to Sentry that are not actually errors we can/should do anything about.


src/domain/app/App.tsx
```diff
@@ -34,8 +36,13 @@ class App extends React.Component<AppProps> {
           this.props.fetchApiTokenError({ message: 'No user found' });
         }
       })
-      .catch(error => this.props.fetchApiTokenError(error));
+      .catch(error => {
+        this.props.fetchApiTokenError(error);
```
midori0507: We should clear oidc localStorage here i think

src/domain/app/App.tsx
```diff
@@ -34,8 +36,13 @@ class App extends React.Component<AppProps> {
...
+        console.log('App.tsx loaduser catch error');
+        console.error(error);
```
midori0507: toast it?

src/domain/api/client.ts
```diff
@@ -15,8 +16,9 @@ export default new ApolloClient({
     } catch (e) {
       // User not authenticated
       // eslint-disable-next-line no-console
+      console.error('Error in client.ts');
       console.error(e);
-      // TODO: add error-handler
+      toast('An error occured. Please try again later');
```
midori0507: We can use `i18n` instance to translate this, locale-based is not necessary but translation should be tracked.

src/domain/app/App.tsx
```diff
@@ -47,6 +54,7 @@ class App extends React.Component<AppProps> {
 
     return (
       <LoadingSpinner isLoading={isLoadingUser}>
+        <ToastContainer></ToastContainer>
```
midori0507: `<ToastContainer />`

src/domain/auth/OidcCallback.tsx
```diff
@@ -14,8 +15,8 @@ function OidcCallback(props: RouteChildrenProps) {
     else props.history.push('/profile');
   };
   const onError = (error: object) => {
-    // TODO: do something about errors
-    props.history.push('/');
+    console.error('OidcCallback error');
```
midori0507: Why still we need this ? Will Sentry catch this?

src/domain/auth/authenticate.ts
```diff
@@ -11,9 +12,17 @@ import { TUNNISTAMO_API_TOKEN_ENDPOINT } from '../api/constants/ApiConstants';
...
 export const loginTunnistamo = (path?: string) => {
-  userManager.signinRedirect(
-    path ? { data: { path: path } } : { data: { path: '/profile' } }
-  );
+  userManager
+    .signinRedirect(
+      path ? { data: { path: path } } : { data: { path: '/profile' } }
+    )
+    .catch(error => {
+      let message = 'Login error';
```
midori0507: 1 generic message here is enough i think

src/domain/registration/form/RegistrationForm.tsx
```diff
@@ -96,12 +99,24 @@ const RegistrationForm: FunctionComponent<Props> = ({
...
-                });
-                history.push('/registration/success');
+                })
+                  .then(result => {
+                    history.push('/registration/success');
+                    console.log(result);
+                  })
+                  .catch(err => {
+                    toast('Failed to submit to server');
+                    if (err instanceof ApolloError) {
```
midori0507: is this follow any practice at all in Kuva ? just toast it i think, this check is not need if `catch` got it

src/domain/app/App.tsx
badrange: At some point, adding a TODO for now

src/domain/app/App.tsx
badrange: I think we can't toast this, it shows up randomly for returning users. Removing the error logs for now.

src/domain/registration/form/RegistrationForm.tsx
badrange: I think some errors, like the user not having a network connection, should not go to Sentry, maybe? It's not like they are bugs we can fix.

src/domain/app/App.tsx
```diff
@@ -34,7 +37,12 @@ class App extends React.Component<AppProps> {
           this.props.fetchApiTokenError({ message: 'No user found' });
         }
       })
-      .catch(error => this.props.fetchApiTokenError(error));
+      .catch(error => {
+        // TODO: Clear oidc local storage when this happens.
+        toast('Failed to log in');
```
midori0507: use i18n

src/domain/auth/authenticate.ts
```diff
@@ -38,7 +50,8 @@ export const authenticateWithBackend = (
 
     dispatch(fetchTokenSuccess(res.data));
   } catch (error) {
+    toast('Failed to get API-token');
```
midori0507: i18n

src/domain/auth/getAuthenticatedUser.ts
```diff
@@ -1,15 +1,28 @@
...
-      resolve(user);
-    } else {
-      reject();
-      // TODO: add error-handler
+    try {
+      const user = await userManager.getUser();
+      if (user) {
+        resolve(user);
+      } else {
+        // TODO: Find out if it is a problem if this happens. The console.error is here to
```
midori0507: non-necessary comments

src/domain/profile/Profile.tsx
```diff
@@ -23,10 +26,18 @@ const Profile: FunctionComponent = () => {
 
   if (loading) return <LoadingSpinner isLoading={true} />;
   if (error || !data || !data.myProfile) {
-    return <Redirect to="/home" />;
+    Sentry.captureException(error);
+    return (
+      <div>
+        <div>Error. Please try again later.</div>
```
midori0507: need translation

src/domain/profile/Profile.tsx
```diff
@@ -23,10 +26,18 @@ const Profile: FunctionComponent = () => {
...
+      </div>
+    );
   } else {
     profile = normalizeProfileData(data);
-    if (profile) dispatch(saveProfile(profile));
+    if (profile) {
+      dispatch(saveProfile(profile));
+      dispatch(resetFormValues());
```
midori0507: dont resetFormValues here, reset when registration form submit success

src/domain/registration/form/RegistrationForm.tsx
```diff
@@ -80,29 +82,32 @@ const RegistrationForm: FunctionComponent<Props> = ({
               setFormIsFilling(false);
               setFormValues(values);
 
+              // TODO: Fix getSupportedChildData so that it works both here
+              // and in ProfileChildrenList.
               const backendSupportChildren = values.children.map(child =>
```
midori0507: revert changes in this file which not related to notification. Keep ticket scope small

badrange: Ok. Good way to test that the toast works, when you are unable to register for the site.

src/domain/profile/Profile.tsx
badrange: ok!

src/domain/auth/authenticate.ts
```diff
@@ -11,9 +14,22 @@ import { TUNNISTAMO_API_TOKEN_ENDPOINT } from '../api/constants/ApiConstants';
...
 export const loginTunnistamo = (path?: string) => {
-  userManager.signinRedirect(
-    path ? { data: { path: path } } : { data: { path: '/profile' } }
-  );
+  userManager
+    .signinRedirect(
+      path ? { data: { path: path } } : { data: { path: '/profile' } }
+    )
+    .catch(error => {
+      if (error.message === 'Network Error') {
```
midori0507: toast only 1 generic one, this is not needed

src/domain/auth/authenticate.ts
```diff
@@ -11,9 +14,22 @@ import { TUNNISTAMO_API_TOKEN_ENDPOINT } from '../api/constants/ApiConstants';
...
+        toast(i18n.t('authentication.networkError.message'), {
+          type: toast.TYPE.ERROR,
+        });
+      } else {
+        toast(i18n.t('authentication.errorMessage'), {
```
midori0507: only this is fine

src/domain/auth/authenticate.ts
badrange: It is not only the toast that is different, we're also not sending sentry errors for network problems

src/domain/auth/authenticate.ts
```diff
@@ -11,9 +14,22 @@ import { TUNNISTAMO_API_TOKEN_ENDPOINT } from '../api/constants/ApiConstants';
...
-  userManager.signinRedirect(
-    path ? { data: { path: path } } : { data: { path: '/profile' } }
-  );
+  userManager
+    .signinRedirect(
+      path ? { data: { path: path } } : { data: { path: '/profile' } }
+    )
+    .catch(error => {
+      if (error.message === 'Network Error') {
+        toast(i18n.t('authentication.networkError.message'), {
```
midori0507: it could be better if you wrote an util
displayErrorNotification= (arg) => toast(...args, { type: toast.TYPE.ERROR })

midori0507: And use it every where instead of repeating that error toast

badrange: Good idea, I will make an issue for this.

PR #66, opened 2019-12-09
KK-192 | Add child profile page

midori0507:


src/domain/child/ChildUtils.ts
```diff
@@ -38,3 +39,15 @@ export const getTranslatedRelationshipOptions = (
     },
   ];
 };
+
+/**
+ * Child data which was submit from form supposed to have same
```
badrange: I think this sentence is a bit clearer than the comment: 

"The child data submitted from the form is supposed to have the same types as the child data received from backend"

midori0507: make sense 👍

PR #67, opened 2019-12-10
KK-123 | Add not eligible warning for child modal

midori0507:
- Add postalCode validation for modal
- Turn off click overlay to hide modal


src/domain/child/modal/ChildFormModal.tsx
```diff
@@ -45,99 +52,131 @@ const ChildFormModal: React.FunctionComponent<ChildFormModalProps> = ({
...
+        {nonEligible ? (
+          <div className={styles.notEligible}>
+            <h3>{t('registration.notEligible.title')}</h3>
+            <p>{t('registration.notEligible.text')}</p>
+            <Icon
+              className={styles.icon}
+              src={personIcon}
+              alt="not eligible icon"
+            />
+            <Button
```
badrange: I think this button should close the modal. If the child is not eligible, we don't want people to go back and "correct" things.

midori0507: was asked to go back to form instead sorry

PR #70, opened 2019-12-10
KK-153 | Use correct relationship format for submit add child mutations

badrange:
This works for me both when registering and adding children. 

Another way to fix this is to open `src/domain/registration/form/partial/ChildFormField.tsx` and change line 92 from 
```name={`children[${childIndex}].relationship.type`}```
to
```name={`children[${childIndex}].relationship`}```

That is a smaller PR but I prefer this one that deletes lines instead.

### DIFF
--- a/src/domain/child/ChildUtils.ts
+++ b/src/domain/child/ChildUtils.ts
@@ -46,8 +46,5 @@ export const getTranslatedRelationshipOptions = (
  * TODO: Fix reducer default data to match backend typing
  */
 export const getSupportedChildData = (child: Child) => {
-  const omited = omit(child, 'homeCity');
-  return Object.assign(omited, {
-    relationship: { type: child.relationship },
-  });
+  return omit(child, 'homeCity');
 };
--- a/src/domain/child/modal/ChildFormModal.tsx
+++ b/src/domain/child/modal/ChildFormModal.tsx
@@ -156,7 +156,7 @@ const ChildFormModal: React.FunctionComponent<ChildFormModalProps> = ({
                 </div>
 
                 <EnhancedInputField
-                  name="relationship"
+                  name="relationship.type"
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: src/domain/child/ChildUtils.ts [hunk 1]
```diff
@@ -46,8 +46,5 @@ export const getTranslatedRelationshipOptions = (
  * TODO: Fix reducer default data to match backend typing
  */
 export const getSupportedChildData = (child: Child) => {
-  const omited = omit(child, 'homeCity');
```
badrange: This is a speling error by the way, should be "omitted"

midori0507: could be yea pls fix

src/domain/child/modal/ChildFormModal.tsx
```diff
@@ -156,7 +156,7 @@ const ChildFormModal: React.FunctionComponent<ChildFormModalProps> = ({
                 </div>
 
                 <EnhancedInputField
-                  name="relationship"
+                  name="relationship.type"
```
midori0507: for the MVP we should keep it as string, if you make this change you gotta change the struc of the reducer data also

midori0507: also typing

src/domain/child/ChildUtils.ts
badrange: I can't correct a spell of ing error in a line of code that gets deleted :-)

src/domain/child/modal/ChildFormModal.tsx
badrange: Which reducer? 
Now I don't understand what you mean - without this change the backend gives us an error when we try to submit data.

PR #71, opened 2019-12-11
KK-135 | KK-137 | Layout for hero and instructions

midori0507:
- Fix language switch and styling along


midori0507:
> I guess the button in the first section isn't supposed to work yet?

Yeah its a placeholder!

PR #72, opened 2019-12-11
Users should not see registration form any more after submitting it

badrange:
Bonus: Remove unused selector from RegistrationForm page.

### DIFF
--- a/src/domain/registration/form/RegistrationForm.tsx
+++ b/src/domain/registration/form/RegistrationForm.tsx
@@ -16,7 +16,6 @@ import submitChildrenAndGuardianMutation from '../mutations/submitChildrenAndGua
 import { resetFormValues, setFormValues } from '../state/RegistrationActions';
 import { RegistrationFormValues } from '../types/RegistrationTypes';
 import { StoreState } from '../../app/types/AppTypes';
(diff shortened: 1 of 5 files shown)


src/domain/registration/form/RegistrationForm.tsx
```diff
@@ -51,6 +53,10 @@ const RegistrationForm: FunctionComponent<Props> = ({
   const history = useHistory();
   const [isOpen, setIsOpen] = useState(false);
 
+  // User can only see form until it has been submitted once. Prevent them
+  // from seeing it again by with use of back button or url hacking.
+  if (userHasProfile) history.push('/');
```
midori0507: Use a return <Redirect /> here instead

src/domain/registration/welcome/Welcome.tsx
```diff
@@ -1,17 +1,45 @@
...
   const history = useHistory();
+
+  const dispatch = useDispatch();
+  let profile;
+
+  if (loading) return <LoadingSpinner isLoading={true} />;
+  if (error || !data || !data.myProfile) {
+    Sentry.captureException(error);
+    return (
+      <div>
```
midori0507: trim this div

PR #74, opened 2019-12-12
Clear user data from browser when logging out

badrange:
At the moment we don't log people out of Tunnistamo when they log out of Kukkuu


src/domain/app/header/userDropdown/UserDropdown.tsx
```diff
@@ -32,11 +32,15 @@ const UserDropdown: React.FunctionComponent = () => {
...
-            // Clear user form data from redux store
-            persistor.flush();
+            // Clear profile (fetched from API)
+            dispatch(clearProfile());
+            // Clear backend auth data
+            dispatch(fetchTokenError());
             // Flush data in redux store and localStorage
-            logoutTunnistamo(location.pathname);
+            persistor.flush();
+            logoutTunnistamo();
```
jimppa: Should user be redirected to some specific page after logout?

src/domain/app/header/userDropdown/UserDropdown.tsx
```diff
@@ -32,11 +32,15 @@ const UserDropdown: React.FunctionComponent = () => {
...
           label: t('authentication.logout.text'),
           onClick: () => {
+            // Clear user form data
             dispatch(resetFormValues());
-            // Clear user form data from redux store
-            persistor.flush();
+            // Clear profile (fetched from API)
+            dispatch(clearProfile());
+            // Clear backend auth data
+            dispatch(fetchTokenError());
```
jimppa: Do you really need to dispatch an error or do you really want to just clear the state?

src/domain/app/header/userDropdown/UserDropdown.tsx
badrange: Good question! If you are on a page that requires authentication, you will end up on the front page automatically, since we have an automatic redirect for the restricted paths. 

Do you think we should add an explicit redirect or just a comment here explaining?

src/domain/app/header/userDropdown/UserDropdown.tsx
badrange: Uhm, this is maybe a badly named action :D

src/domain/app/header/userDropdown/UserDropdown.tsx
jimppa: Ah, I guess it's fine like this if the router keeps track of that.

src/domain/app/header/userDropdown/UserDropdown.tsx
jimppa: It looks like the action is used for just that intent (token fetch error), I think it would be best to create a new action + reducer for resetting the backend auth state.

badrange: Better now?

src/domain/auth/state/BackendAuthenticationReducer.ts
```diff
@@ -26,4 +26,10 @@ export default createReducer(defaultApiAuthenticationData, {
       apiToken: null,
       errors: action.payload,
     }),
+  [API_AUTHENTICATION_ACTIONS.RESET_BACKEND_AUTHENTICATION]: (state, action) =>
+    Object.assign({}, state, {
+      isFetchingToken: false,
+      isAuthenticated: false,
+      apiToken: null,
```
jimppa: I would reset errors as well

src/domain/auth/authenticate.ts
```diff
@@ -32,8 +32,14 @@ export const loginTunnistamo = (path?: string) => {
     });
 };
 
-export const logoutTunnistamo = (path?: string) => {
-  userManager.signoutRedirect(path ? { data: { path: path } } : {});
+// TODO Decide how logout should actually work - do we really want
+// to log people out of Tunnistamo itself?
```
midori0507: Why dont we?

src/domain/auth/state/BackendAuthenticationReducer.ts
midori0507: Fixed that

src/domain/auth/authenticate.ts
badrange: When you use FB-login to log in to some random service, you don't expect to be logged out of fb when you log out of that random service. This use case is a bit similar. 

We still need to have a conversation about this with the other teams that use Tunnistamo.

PR #77, opened 2019-12-12
Tweaks to design

badrange:
I looked at the front page with Anniina, and added a few small improvements. The h1, h2 and h3 changes are accessibility improvements.


src/domain/home/hero/hero.module.scss
```diff
@@ -15,6 +15,9 @@
     background-size: 100% auto;
     display: grid;
     padding-bottom: $largeMargin;
+    @include respond-above(lg) {
+        min-height: 35vh;
```
midori0507: We might need to increase it a bit, 40 and 45 is better i think

badrange: Looked at it with Anniina, I increased the value in the developer console until she said "stop, now it looks good"

midori0507: Ok

PR #80, opened 2019-12-13
If logging in without a profile, be redirected to the front page

badrange:


src/domain/profile/Profile.tsx
```diff
@@ -24,12 +24,15 @@ const Profile: FunctionComponent = () => {
...
-    return (
-      <div>
-        <div>{t('api.errorMessage')}</div>
-      </div>
-    );
+    if (error) {
+      Sentry.captureException(error);
+      return (
+        <div>
+          <div>{t('api.errorMessage')}</div>
```
midori0507: Replace div with PageWrapper otherwise that text will have no margin

src/domain/profile/profile.module.scss
```diff
@@ -0,0 +1,6 @@
+.profile {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    align-self: center;
```
midori0507: You dont need this line

badrange: Removed

PR #89, opened 2019-12-19
KK-171 | Accessibility improvements

badrange:
It should be possible to navigate the user interface using the keyboard
We should indicate the role of buttons to users with screen readers


src/domain/app/header/userMenu/UserMenu.tsx
```diff
@@ -18,7 +18,7 @@ const UserMenu: React.FunctionComponent<UserMenuProps> = ({ options }) => {
       {options.map((option, index) => {
         return (
           <li key={index} onClick={() => option.onClick && option.onClick()}>
-            {option.label}
+            <button>{option.label}</button>
```
badrange: I'm pretty sure it would be more "correct" to have a link here instead of a button, but that improvement requires a refactoring of the whole user menu. I propose we spend time on that if someone complains, the site should be usable as it is!

src/common/translation/i18n/en.json
```diff
@@ -337,13 +345,21 @@
...
-  }
+  },
+  "email": {
+    "title": "Welcome to the Culture Kids programme!"
+  },
+  "email ": {
+    "heading": "Welcome to the Culture Kids programme!",
+    "text": "Welcome!"
+  },
+  " Please note! You cannot reply to this message.": ""
```
tuomas777: I don't think these email texts should be here (?)

badrange: If they are in the sheet, they will automatically also come here, we don't have a filter in the script that fetches the translations. I don't think it hurts the project to have them - but they can be moved to their "own" tab in the spreadsheet at some point.

PR #92, opened 2019-12-23
Feature/kk 158 on the front page the background colors should go from edge to edge

ironroger:


src/domain/app/layout/Container.tsx
```diff
@@ -2,12 +2,36 @@ import React, { FunctionComponent } from 'react';
...
+
+const stringStartsWith = (targetString: string, searchString: string) => {
+  return targetString.substring(0, searchString.length) === searchString;
+};
+
+const Container: FunctionComponent<{
+  children?: any;
+  className?: string;
+}> = ({ children, className }) => {
+  const isHome =
```
jimppa: This component shouldn't know anything of where it's used. Since the purpose of this PR is to set the width only on Home-page i would suggest you do the CSS-fixes there instead (`src/domain/home`) and reset everything that was made to this component. Conveniently this component already supports passing className from parent components, use that for styling this component differently.

PR #94, opened 2020-01-02
KK-159 | Front page hero header text should not be on top of kids change kids image

ironroger:


src/assets/styles/layout.scss
```diff
@@ -109,9 +109,29 @@ $breakpoints: (
...
     grid-template-columns: 1fr minmax(auto, $containerFormMaxWidth) 1fr;
-    
+
     > * {
         grid-column: 2;
     }
+  }
+}
+
+@mixin heroContainer() {
```
jimppa: So the resulting media queries will be targeted at (max-width 1199px), (min-width 992px) which doesn't make much sense unless I'm missing something here. Wouldn't it be better to write this in a mobile-first manner?
```
@mixin heroContainer() {
  display: grid;
  grid-template-columns: 1fr 10fr 1fr;
  > * {
    grid-column: 2;
  }
  @include respond-above(md) {
    grid-template-columns: 1fr minmax(auto, $containerFormMaxWidth) 1fr;
  }
}
```

src/domain/home/hero/hero.module.scss
```diff
@@ -5,29 +5,44 @@
...
+        background-image: url('../../../assets/icons/svg/Culture_kids_background@2x.png');
+        background-repeat: repeat-x;
+        background-position: center bottom -2px;
+        background-size: 100% 321px;
     }
 }
 
+.kidsImage {
+    grid-column: 2;
+    @include respond-above(md) {
```
jimppa: These media-queries don't make much sense. (min-width 992px), then next one (max-width 1199px). Instead of using max-width media queries, just make that the default style, then from a certain min-width override what's necessary. Max-width media queries are very rarely needed.

PR #97, opened 2020-01-09
Fix custom birthdate validation not getting translated

midori0507:
@Zeukkari : You can ignore changes made to json files, it happened because i sort the Google translation sheet

`validateDate` util return raw string which is not translated

### DIFF
--- a/src/domain/home/form/partial/BirthdateFormField.tsx
+++ b/src/domain/home/form/partial/BirthdateFormField.tsx
@@ -12,7 +12,7 @@ const BirthdateFormField: FunctionComponent<FieldArrayRenderProps> = ({
   form: { errors, touched },
 }) => {
   const fieldTouched = getIn(touched, 'child.birthdate.day');
-  const error = errors['childBirthdate'];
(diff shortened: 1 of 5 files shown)


src/domain/home/form/partial/BirthdateFormField.tsx
```diff
@@ -12,7 +12,7 @@ const BirthdateFormField: FunctionComponent<FieldArrayRenderProps> = ({
   form: { errors, touched },
 }) => {
   const fieldTouched = getIn(touched, 'child.birthdate.day');
-  const error = errors['childBirthdate'];
+  const error = getIn(errors, 'childBirthdate');
```
Zeukkari: this could be done with plain typescript 3.7:
```
const error = errors?.childBirthdate
```

midori0507: that's nice, but our ts version is lower than 3.7 
Will take this in use after upgrade, would be nice to remove lot of hacky nullcheck 👍

PR #98, opened 2020-01-09
KK-188 | use maxLength to limit amount of char in input number

midori0507:
HTML number input doesnt respect `min` and `max` property. 
Manually give custom `maxLength` limit to number input

### DIFF
--- a/src/common/components/form/fields/input/NumberInputField.tsx
+++ b/src/common/components/form/fields/input/NumberInputField.tsx
@@ -1,0 +1,43 @@
+import React, { KeyboardEvent } from 'react';
+import { FieldProps, getIn } from 'formik';
+import classnames from 'classnames';
+import { useTranslation } from 'react-i18next';
+
+import Input from '../../../input/Input';
+import styles from './inputField.module.scss';
+
+interface NumberInputField extends FieldProps {
+  id: string;
+  label: string;
+  maxLength: number;
+}
+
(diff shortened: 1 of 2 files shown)


src/common/components/form/fields/input/NumberInputField.tsx
```diff
@@ -0,0 +1,43 @@
...
+import Input from '../../../input/Input';
+import styles from './inputField.module.scss';
+
+interface NumberInputField extends FieldProps {
+  id: string;
+  label: string;
+  maxLength: number;
+}
+
+const NumberInputField: React.ComponentType<NumberInputField> = ({
```
midori0507: I dont want to put more logic to single InputField component so i make new component which look almost same for number input only. Sorry for duplication

src/common/components/form/fields/input/NumberInputField.tsx
```diff
@@ -0,0 +1,43 @@
...
+  maxLength,
+  ...rest
+}) => {
+  const error = getIn(errors, field.name);
+  const touch = getIn(touched, field.name);
+  const inputError = touch && error ? error : null;
+
+  const onKeyPress = (e: KeyboardEvent<HTMLInputElement>) => {
+    if (maxLength && field.value.toString().length >= maxLength) {
+      e.preventDefault();
```
midori0507: Cancel event bubbling when limit was reach

PR #104, opened 2020-01-14
KK-191 | Make static homepage for displaying temporary under-construction page. **DO NOT MERGE**

midori0507:
This is only for temp deployment. 
Used 1 time only, so it wont be a part of `develop`
**DONT MERGE**

Will remove after live deployment.


midori0507:
Review done, close branch and ready to deploy

PR #112, opened 2020-01-20
KK-203 | Use real typing than overrided types for Profile

midori0507:
I think bumping CRA make the standard go higher, so lets solve the mutating problem by just not mutating at all 😄


badrange:
Hang on, I think it is connected to using the back button somewhere?

midori0507:
@badrange Yeah i think we gotta manually update profile cache after mutation

src/domain/profile/Profile.tsx
```diff
@@ -35,12 +31,8 @@ const Profile: FunctionComponent = () => {
       </PageWrapper>
     );
   }
-  // userHasProfile selector saves us when Apollo cache gives a wrong answer.
```
badrange: I managed to enter a weird state when pressing the back button right after submitting the first page form and getting logged in.

midori0507: But this logic is a bit wrong, gotta fix it another way.
Can you make step by step to reproduce?

midori0507: data that fetched should always be the single source of truth, not cached one

badrange: Steps to reproduce - and it might be that this is unrelated to this change:

1. delete your profile
2. Submit front page form (so that you're logged in again)
3. press back button
4. try to continue with registration

badrange: Based on the superficial testing we did, I think we can mark this one as resolved - and I still think it makes sense to look for any issues on the profile page. Hopefully the error doesn't happen after removing the normalizeProfileData function.

PR #114, opened 2020-01-20
KK-197 | Add Accessibility statement

badrange:


src/domain/accessibilityStatement/accessibilityStatement.module.scss
```diff
@@ -0,0 +1,19 @@
...
+@import 'variables';
+
+.accessibilityStatement {
+  margin-bottom: $largeMargin;
+  @include respond-below(md) {
+    margin: 0 $baseMargin $largeMargin $baseMargin;
+  }
+
+  h1 {
+    font-size: var(--hds-text-xxl);
```
midori0507: We should not tho, even its not fitting our eyes, to keep consistency of HDS.
If h1 was too big, change it to h2 then h3 h4..

badrange: That is also wrong, we should keep the semantics of the content of the document. Especially for h1. I'll remove these, then we can negotiate with the content creator.

PR #120, opened 2020-01-23
KK-214 | Add translatable title&meta-description and image for social sharing

badrange:
This also 
* fixes an accessibility issue where html language was always en no matter content language
* fixes a bug where the page title on the front page was always the same untranslated `Kulttuurin Kummilapset` no matter which language you were looking at.


public/index.html
```diff
@@ -1,10 +1,10 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="fi">
```
badrange: I set this to fi because I believe 80+% of our users are Finnish
Not sure if there will be an actual consequence of this
Site title is also in Finnish by default

badrange:
image

PR #124, opened 2020-01-27
Improve performance of front end

badrange:
image

* Change kids image from 2MB png to 364K jpg
* Set a long expire time for files
* Do not load Noto Serif font from google fonts - it is not in use


.prod/nginx.conf
```diff
@@ -11,6 +11,7 @@ server {
     }
 
     location / {
+        expires max;
```
midori0507: Are you sure sure? I dont have knowledge in nginx much but if you know what you are doing then its fine 🤔

badrange: webpack gives each file in the bundle a unique name, so this should be safe

PR #127, opened 2020-01-28
KK-214 | Add more og tags and twitter card

badrange:


public/index.html
```diff
@@ -1,14 +1,21 @@
 <!DOCTYPE html>
-<html lang="fi">
+<html prefix="og: http://ogp.me/ns#" lang="fi">
```
midori0507: Where is that link heading to? Can we host that xml file ourselve?

badrange: This is taken from the Open Graph protocol site: https://ogp.me/

PR #130, opened 2020-01-29
KK-162 | Extra style reviews

midori0507:


src/domain/app/layout/pageWrapper.module.scss
```diff
@@ -1,3 +1,7 @@
 .pageWrapper {
   display: grid;
 }
+
+.defaultContainer {
+  padding-bottom: 5rem;
```
midori0507: This put extra 3rem padding to bottom of container to separate 2rem of footer Koros, this can be reduce to 3rem when HDS Koros is in used

PR #140, opened 2020-01-31
Use verb form of "Log in"

hugovk:
"Login" is a noun, use "Log in" as a verb instead.

image

See:

* https://www.facebook.com

image

* https://grammarist.com/spelling/log-in-login/
* http://notaverb.com/login

PR #143, opened 2020-01-31
Simplify text

hugovk:
I think this reads a little better.

The original sounds a bit more like a direct Finnish translation.

Shorter is usually better online, people scan rather than reading text.

### DIFF
--- a/src/common/translation/i18n/en.json
+++ b/src/common/translation/i18n/en.json
@@ -12,10 +12,10 @@
       "message": "An error occurred. Please try logging in again. "
     },
     "login": {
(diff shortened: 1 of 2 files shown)


badrange:
Thank you for this suggestion and the ones from #140 and #144. We have now updated the spreadsheet that hosts the translations with your improvements. 

If you merge or rebase develop into this branch and run `yarn update-translations` you'll get a new en.json file that you can add to this pull request.

hugovk:
Done!

badrange:
Brilliant!

PR #151, opened 2020-02-03
KK-222 | Display user session timeout prompt when token expired.

midori0507:
- Remove apollo-boost. Split default apollo-boost into multiple link for better control and add apollo-link-error as global error interceptor. 
- Improve auth flow, logout user when error
- Prevent over-fetch apiToken from backend when route changed


src/domain/auth/state/AuthenticationReducers.ts
```diff
@@ -1,9 +1,17 @@
...
 import { combineReducers } from 'redux';
+import storage from 'redux-persist/lib/storage';
+import { persistReducer } from 'redux-persist';
 
 import backendAuthenticationReducer from './BackendAuthenticationReducer';
 
+const backendPersistConfig = {
+  key: 'backendPersistor',
+  storage,
+  whitelist: ['apiToken'],
```
midori0507: persist the apiToken, prevent overfetching when navigating around app.
When token expired -> apollo error gonna catch the error -> display prompt -> log user out.

src/domain/app/state/ui/UIReducers.ts
```diff
@@ -0,0 +1,19 @@
+import { createReducer } from '@reduxjs/toolkit';
```
midori0507: global prompt display. Its kinda long but its better than hook h@ck

src/domain/app/state/AppStore.ts
```diff
@@ -20,6 +22,8 @@ const store: Store = configureStore({
   reducer: persistedReducer,
 });
 
+loadUser(store, userManager);
```
midori0507: loadUser in place where it should stay

src/domain/app/sessionPrompt/SessionPrompt.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+const SessionPrompt: React.FunctionComponent<{ isOpen: boolean }> = ({
+  isOpen = false,
+}) => {
+  const dispatch = useDispatch();
+
+  const flushAuthenticationData = () => {
+    dispatch(closeExpiredSessionPrompt());
+
+    dispatch(resetBackendAuthentication());
```
midori0507: Flush token

src/domain/app/App.tsx
```diff
@@ -12,102 +9,71 @@ import NotEligible from '../registration/notEligible/NotEligible';
...
-            </PrivateRoute>
-          )}
-          <PrivateRoute exact path={`/${locale}/registration/success`}>
-            <Welcome />
-          </PrivateRoute>
+  useEffect(() => {
+    if (!apiToken && user?.access_token) {
+      dispatch(authenticateWithBackend(user.access_token));
+    }
+  }, [apiToken, dispatch, user]);
```
midori0507: This is better than componentDidMount, which only do token exchange when there are no token. 
In future when silent-renew work, we can remove the `!apiToken`, it will auto get new token when user data change.

(useEffect subscribe for apiToken / user changes)

badrange: What if you add a comment about that future option?

midori0507: Sure things

src/common/translation/i18n/en.json
```diff
@@ -22,6 +22,13 @@
     },
     "redirect": {
       "text": "Redirecting..."
+    },
+    "session": {
+      "expired": {
+        "message": "Your session have expired. <br>In order to continue with application <br> you will be redirect to homepage and start again. <br>All your current progress will be saved.",
```
badrange: I think a simpler error message would be nice. Maybe just: "Your session has expired. Go to the front page."

This might also show up when people are finished registering and are signing up for events etc, so "Continue registration" is pretty confusing?

midori0507: Can you edit the sheet?

src/domain/api/client.ts
```diff
@@ -1,25 +1,46 @@
...
+
+const errorLink = onError(({ graphQLErrors, networkError }) => {
+  if (graphQLErrors) {
+    toast('authentication.loadUserError.message');
+  }
+
+  if (networkError) toast(i18n.t('authentication.network.error'));
+
+  store.dispatch(showExpiredSessionPrompt());
+  // Now show session timeout for all kind of error.
```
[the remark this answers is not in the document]
midori0507: might toast a generic message only i think, we can improve the handler when we have the statusCode from backend.

src/domain/app/sessionPrompt/SessionPrompt.tsx
```diff
@@ -8,17 +8,30 @@ import { closeExpiredSessionPrompt } from '../state/ui/UIActions';
...
   const dispatch = useDispatch();
 
   const flushAuthenticationData = () => {
+    // Close prompt
     dispatch(closeExpiredSessionPrompt());
 
+    // Clear api token
     dispatch(resetBackendAuthentication());
 
+    // Clear profile (fetched from API)
```
badrange: Now we have two places where these logout operations are done, with a small difference. Wonder if it would be better to create a function for this with some parameter? 

A future dev might be confused or annoyed if they for some reason need to add one more reducer (or something) during logout, and then end up adding it two places..

src/domain/auth/state/AuthenticationUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+import { clearProfile } from '../../profile/state/ProfileActions';
+import { resetBackendAuthentication } from './BackendAuthenticationActions';
+import client from '../../api/client';
+import { resetUIState } from '../../app/state/ui/UIActions';
+
+type FlushStateOptions = {
+  keepUserFormData?: boolean;
+};
+
+export const flushAllState = (options: FlushStateOptions) => {
```
badrange: Would be great with a comment here about logging out so that people can find it if they search for logout or logOut or log out in our codebase

midori0507: No i dont include log out in this function, so no need to explain here

PR #152, opened 2020-02-04
add silent renew outside React router

midori0507:


midori0507:
Oh boi this doesnt work, might need better routing

PR #154, opened 2020-02-05
KK-215 | Add usage analytics

badrange:


src/common/components/dropdown/Dropdown.tsx
```diff
@@ -42,13 +43,16 @@ const Dropdown: React.FunctionComponent<DropdownProps> = ({
 
   const ariaLabel = isOpen
     ? t('common.menu.closeMenuText')
+    : options.length === 1
```
midori0507: where is this idea come from?

midori0507: The double chaining is really bad for reading just saying

badrange: I know it is bad, I was too lazy to make anything better right now, maybe we have too little oxygen in the office. I will come up with something better tomorrow unless you have a suggestion out of hand. 

The problem is that the aria-label & aria-expanded info is totally wrong on the user/drop down menu on desktop view. It always stays as `Open menu` even if it should be `Log in`.

midori0507: I can commit straight to make it more readable, take some rest.

badrange: Be my guest! 😻

PR #163, opened 2020-02-11
KK-123 | Add edit and delete child mutation

midori0507:


src/common/translation/i18n/en.json
```diff
@@ -45,7 +45,8 @@
           "text": "Add child"
         },
         "delete": {
-          "label": "Remove"
+          "label": "Remove",
+          "text": "Delete child"
```
badrange: This should read Remove child profile like profile┊child┊detail┊delete┊text

src/domain/child/modal/ChildFormModal.tsx
```diff
@@ -198,6 +213,16 @@ const ChildFormModal: React.FunctionComponent<ChildFormModalProps> = ({
                 >
                   {t('child.form.modal.add.label')}
```
badrange: This is a bug from before :) It shouldn't show the add label when you are editing!

midori0507: Fixed

PR #166, opened 2020-02-12
Add max-width for pages with text in them, fix footer links

badrange:


src/assets/styles/variables.scss
```diff
@@ -49,3 +49,6 @@ $containerPaddingBottom: 4rem;
 // Line height
 $baseLineHeight: 1.2;
 $largeLineHeight: 1.6;
+
+// Maxwidth for text paragraphs
+$textParagraphMaxWidth: 50rem;
```
midori0507: https://i.dlpng.com/static/png/5480524_thumb.png
how come you come up with this idea??

src/domain/app/footer/Footer.tsx
```diff
@@ -16,9 +17,15 @@ const Footer: FunctionComponent = () => {
             <p>{t('footer.copyrightText')}</p>
           </div>
           <div className={styles.links}>
-            <a href="/accessibility">{t('accessibilityStatement.title')}</a> •{' '}
-            <a href="/terms">{t('termsOfService.title')}</a> •{' '}
-            <a href={t('descriptionOfTheFile.url')}>
+            <Link to="/accessibility#start">
+              {t('accessibilityStatement.title')}
```
midori0507: The `{  }` is quite bad tho, can we replace it with padding / margin instead?

src/domain/accessibilityStatement/accessibilityStatement.module.scss
```diff
@@ -3,6 +3,7 @@
 
 .accessibilityStatement {
   margin-bottom: $largeMargin;
+  max-width: $textParagraphMaxWidth;
```
midori0507: This look very duct tape

src/assets/styles/variables.scss
badrange: Anniina. It is recommended max-line-length for text.

src/domain/app/footer/Footer.tsx
badrange: It is a space character. We had it before and I'd like to keep it :-)

src/domain/accessibilityStatement/accessibilityStatement.module.scss
badrange: Any idea on how to improve on this? The issue is that the lines should not be too long, it is difficult to read text which has very long lines.

src/domain/app/footer/Footer.tsx
midori0507: alrigthy then

PR #174, opened 2020-02-17
KK-261 | Enable silent renewing of auth tokens

badrange:
This implementation is inspired by https://github.com/City-of-Helsinki/berth-reservations-admin/pull/166/files


midori0507:
@badrange : The idea only make sense if the PR actually revive the silent-renew feature. 
Add more config but silent renew still b0rken is quite unnecessary

PR #175, opened 2020-02-17
Bumping dependencies

midori0507:
Latest CRA, node-sass and apollo
Current security happen from Apollo basically, have yarn as dependency


badrange:
`yarn && yarn start` worked after I `rm -rf node_modules `

PR #180, opened 2020-02-18
Send environment info to sentry, only track analytics from production

badrange:


src/domain/app/BrowserApp.tsx
```diff
@@ -33,6 +33,13 @@ const instance = createInstance({
   siteId: 56,
 });
 
+// Pretend that we require consent to process analytics data without asking for it on
+// non-production environments to avoid having analytics from dev environments in Matomo.
+// https://developer.matomo.org/guides/tracking-javascript-guide#step-1-require-consent
+if (process.env.REACT_APP_ENVIRONMENT !== 'production') {
```
midori0507: why dont we just use NODE_ENV?

badrange: NODE_ENV is ALWAYS production except for when you're running it on port 3000 on your local machine

PR #181, opened 2020-02-18
KK-250 | Ensure that login page is in user's chosen language

### LINKED ISSUE #146: Login page in Finnish when using English or Swedish site
1. Start from English (or Swedish) site
2. Enter details and click "Register"

image

Expected result: login page is English (or Swedish)

Actual result: login page is Finnish

image

badrange:
This should fix #146 - it does require that Tunnistamo is configured correctly.

### DIFF
--- a/src/domain/auth/authenticate.ts
+++ b/src/domain/auth/authenticate.ts
@@ -12,12 +12,15 @@ import {
 } from './state/BackendAuthenticationActions';
 import { TUNNISTAMO_API_TOKEN_ENDPOINT } from '../api/constants/ApiConstants';
 import { BackendTokenResponse } from './types/BackendAuthenticationTypes';
+import { getCurrentLanguage } from '../../common/translation/TranslationUtils';
 
 export const loginTunnistamo = (path?: string) => {
   userManager
-    .signinRedirect(
-      path ? { data: { path: path } } : { data: { path: '/profile' } }
-    )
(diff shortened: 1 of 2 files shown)


src/domain/auth/authenticate.ts
```diff
@@ -12,12 +12,15 @@ import {
...
+import { getCurrentLanguage } from '../../common/translation/TranslationUtils';
 
 export const loginTunnistamo = (path?: string) => {
   userManager
-    .signinRedirect(
-      path ? { data: { path: path } } : { data: { path: '/profile' } }
-    )
+    .signinRedirect({
+      data: { path: path || 'profile' },
+      /* eslint-disable @typescript-eslint/camelcase */
```
midori0507: Imported I18n instance doesnt reflect user selected language correctly.
Only i18n from component is.

badrange: We can only use i18n from component if we are working with a component? I see no FC component around here?

I tested my implementation, and it works correctly also when changing language.

midori0507: Yeah just worry if you need exact correct language, lgtm

badrange: One more possible corner case bug in here :D 
But it is at least in the same file. And the risk is low and impact negligible.

PR #186, opened 2020-02-21
KK-299 | RFC: Event page version 0.0.1

badrange:
To test, go directly to the url of an event (you need to figure out the uuid somehow) like this:
http://localhost:3000/fi/event/RXZlbnROb2RlOjE=

It looks something like this: 
image

Note that having TODOs in here is OK, we will resolve them in KK-300.

### DIFF
--- a/src/domain/app/App.tsx
+++ b/src/domain/app/App.tsx
@@ -26,7 +26,7 @@ import {
   fetchTokenError,
 } from '../auth/state/BackendAuthenticationActions';
 import ProfileRoute from '../profile/route/ProfileRoute';
-import Event from '../event/Event';
+import EventRoute from '../event/route/EventRoute';
 
 const App: React.FunctionComponent = props => {
   const isLoadingUser = useSelector(isLoadingUserSelector);
@@ -97,7 +97,7 @@ const App: React.FunctionComponent = props => {
           // we're waiting for KK-277.
         }
         <PrivateRoute path={`/${locale}/event`}>
-          <Event />
+          <EventRoute />
         </PrivateRoute>
 
         {userHasProfile && <Redirect to={`/${locale}/profile`} />}
--- a/src/domain/app/App.tsx
+++ b/src/domain/app/App.tsx
@@ -92,10 +92,6 @@ const App: React.FunctionComponent = props => {
           <ProfileRoute />
         </PrivateRoute>
 
-        {
-          // TODO: The route should probably be /profile/child/${child.id}/events/${event.id}
-          // we're waiting for KK-277.
-        }
         <PrivateRoute path={`/${locale}/event`}>
           <EventRoute />
         </PrivateRoute>
--- a/src/domain/app/App.tsx
+++ b/src/domain/app/App.tsx
@@ -26,6 +26,7 @@ import {
   fetchTokenError,
 } from '../auth/state/BackendAuthenticationActions';
 import ProfileRoute from '../profile/route/ProfileRoute';
+import Event from '../event/Event';
 
 const App: React.FunctionComponent = props => {
   const isLoadingUser = useSelector(isLoadingUserSelector);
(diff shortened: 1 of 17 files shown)


src/domain/event/Event.tsx
```diff
@@ -0,0 +1,122 @@
...
+  };
+
+  // TODO: Get from API (all todo's are pending that)
+  const selectOptions = [
+    { value: 1, label: 'one' },
+    { value: 2, label: 'two' },
+  ];
+  // TODO:
+  const bgImageUrl =
+    'https://edit.myhelsinki.fi/sites/default/files/styles/hero_image/public/2019-12/flow_festival_2019_friday_c_petri_anttila_mg_0397.jpg';
```
midori0507: lol

src/domain/app/App.tsx
```diff
@@ -91,6 +92,14 @@ const App: React.FunctionComponent = props => {
           <ProfileRoute />
         </PrivateRoute>
 
+        {
+          // TODO: The route should probably be /profile/child/${child.id}/events/${event.id}
```
midori0507: This routing is ridiculous, i think /event/eventid will be neat

src/domain/app/header/navbar/navbar.module.scss
```diff
@@ -4,9 +4,6 @@
 @import 'mixins';
 @import 'variables';
 
-$logoHeight: 2.5rem;
```
midori0507: Why? you reuse it or something?

src/domain/app/App.tsx
```diff
@@ -91,6 +92,14 @@ const App: React.FunctionComponent = props => {
...
+          // we're waiting for KK-277.
+        }
+        <PrivateRoute path={`/${locale}/event`}>
+          <Event />
```
midori0507: can user who doesnt have profile able to see this?

src/domain/app/App.tsx
badrange: Fair enough!

src/domain/app/header/navbar/navbar.module.scss
badrange: I figured it could be nice to have those two variables together in the same place. A future dev might be confused as to why we have them in different files / have trouble finding them. 

Do you think I should move the height one back?

src/domain/app/App.tsx
badrange: Based on what I know this should be a private route. Only logged in people can be invited and sign up for occurrences.

src/domain/event/Event.tsx
badrange: Backend now returns the fully qualified url

src/domain/event/EventOccurrence.tsx
```diff
@@ -0,0 +1,39 @@
...
+  // TODO: Ensure that date shows in correct locale.
+  const date = formatTime(newMoment(occurrence.time), 'dd l');
+  const time = formatTime(newMoment(occurrence.time), 'hh:mm');
+
+  return (
+    <tr className={styles.occurrence}>
+      <td>{date}</td>
+      <td>{time}</td>
+      <td>
+        {occurrence.venue.name} {occurrence.venue.description}
```
jpsalo: Should venue's description not be here?

src/domain/event/Event.tsx
```diff
@@ -0,0 +1,147 @@
...
+              </div>
+              <div className={styles.signup}>
+                <Formik
+                  initialValues={initialValues}
+                  onSubmit={handleSubmit}
+                  validate={(values: SignupValues) => {
+                    handleSubmit(values);
+                  }}
+                >
+                  {({ handleSubmit, handleChange }) => {
```
jpsalo: Is `handleChange` TODO?

src/domain/event/EventOccurrence.tsx
badrange: I actually don't know what should be there, we'll find out pretty soon..

src/domain/event/Event.tsx
badrange: Yes it is TODO..

PR #188, opened 2020-02-27
KK-299 | Add tests for event components

badrange:


src/domain/event/Event.tsx
```diff
@@ -86,7 +86,7 @@ const Event: FunctionComponent = () => {
         </div>
       </div>
 
-      <PageWrapper className={styles.wrapper} title={'event'}>
+      <PageWrapper className={styles.wrapper} title={data?.event?.name || ''}>
```
midori0507: can we have fallback to maybe  `event detail` instead?

badrange: I actually think we should return something else entirely if data.event is empty, for example not found..

PR #189, opened 2020-02-28
KK-277 | Child page event listing

jpsalo:
Draft notes:
- [ ] There's a bunch of TODOs which don't affect the functionality but will be resolved before merge
- [ ] Image is mercilessly resized to 200x200 for now
- [x] Typings require reiteration
- [x] It's mobile first but not mobile-optimized
- [ ] event duration determines occurrence duration - now it's hard-coded to 30 minutes
- [ ] calendar icon is missing
- [ ] enrolments return past events but it will be fixed in the backend
- [ ] tests for empty list cases are missing

localhost_3000_fi_profile_child_Q2hpbGROb2RlOjliODJiOTQxLWZlMWMtNGQwMy1hMTU4LWZmOTM5YmI3NWIwNw== (1)


.env.development
```diff
@@ -1,2 +1,2 @@
-REACT_APP_API_URI="https://kukkuu.test.kuva.hel.ninja/graphql"
```
midori0507: Uhm maybe no, keep it currently

package.json
```diff
@@ -69,6 +69,7 @@
     "test": "react-scripts test --watchAll=false",
     "test:debug": "react-scripts --inspect-brk test --runInBand --no-cache"
   },
+  "proxy": "http://localhost:8081",
```
midori0507: why this one defined here ?

src/common/components/card/card.module.scss
```diff
@@ -0,0 +1,90 @@
...
+    align-items: center;
+    margin: 0;
+    padding: 0;
+  }
+  background: $white;
+  & + & {
+    margin-top: 20px;
+  }
+  margin: 0 -1rem;
+  padding: 1rem 1rem;
```
midori0507: just padding: 1rem  ? 😄

src/common/components/card/card.module.scss
```diff
@@ -0,0 +1,90 @@
...
+  }
+  >:last-child {
+    margin-bottom: 0;
+  }
+  margin: 20px 0;
+}
+
+.end {
+  @include respond-above(sm) {
+    margin-right: 20px;
```
midori0507: be careful with duplicated free value like this, we discussed yesterday

src/domain/child/queries/ChildQueries.ts
```diff
@@ -40,6 +40,63 @@ export const childByIdQuery = gql`
...
+              }
+            }
+          }
+        }
+      }
+      availableEvents {
+        edges {
+          node {
+            id
+            translations {
```
badrange: you can get the name directly here if you have an up to date backend
No need to fiddle with translations!

src/common/components/card/Card.tsx
```diff
@@ -0,0 +1,67 @@
...
+  primaryActionText,
+  title,
+}) => {
+  const { t } = useTranslation();
+
+  return (
+    <div className={styles.wrapper}>
+      <div className={styles.start}>
+        {/* TODO: alt, size */}
+        <img src={image} alt={'alt'} width="200" height="200" />
```
badrange: For mobile I don't think you need to define the width, you can just let it fill the screen? 

And if you don't define height either you can allow the editors to choose the image aspect ratio (not sure if that is a good idea)

.env.development
```diff
@@ -1,2 +1,2 @@
...
+REACT_APP_API_URI="ttps://kukkuu.test.kuva.hel.ninja/graphql"
```
badrange: I think an H is missing :D

PR #190, opened 2020-02-28
KK-299 | Add special mobile display of event occurrence list

badrange:


src/domain/event/eventOccurrence.module.scss
```diff
@@ -8,16 +8,46 @@
...
+    grid-column: 1 / span 3;
+  }
+  td.occurrenceDate {
+    grid-row: 2;
+    grid-column: 1 / span 2;
+  }
+  td.occurrenceTime {
+    grid-row: 2;
+    grid-column: 2 / span 2;
+  }
```
jpsalo: Date and time starts to overlap at around 400px width. Is it something we need to consider/support?
localhost_3000_fi_event_RXZlbnROb2RlOjE=

badrange: Do any device exist that have a 400px screen width?

badrange: Let me see if we can find a better solution

badrange: Does it work better now?

PR #193, opened 2020-03-03
KK-278 | Show a QR code instead of event image for upcoming events

badrange:
* There is a requirement to show QR on mobile and image on desktop
* Resolve image-alt todo in card component


badrange:
We might be able to get away with showing a static image of a QR code here, but oh well.

badrange:
> It works as expected but please try to keep `<Card />` as primitive as possible and implement the logic to set QR code as a `<Card image={image OR qr} />`

That was a good point. Turns out that we can't use image OR qr since we're supposed to show QR code on mobile and image on desktop. 

I also noticed that the qr library I used increased the package size by a lot, I found another (used by youth-membership-ui project) and added that instead.

src/common/components/card/Card.tsx
```diff
@@ -9,30 +11,37 @@ import Icon from '../icon/Icon';
...
   primaryAction?: () => void;
   primaryActionText?: string;
+  qr?: string;
   title: string;
 }
 
 const Card: FunctionComponent<CardProps> = ({
   action,
   actionText,
+  alt = '',
```
badrange: It should be fine to have an empty alt text: If the image has no information / is only decorative it is better to send a blank alt text to screen readers.

PR #194, opened 2020-03-04
KK-331 | Child page event list tests

jpsalo:
Testin' and Fixin'


src/domain/profile/events/ProfileEvents.tsx
```diff
@@ -10,7 +10,17 @@ interface ProfileEventsProps {
 
 const ProfileEvents: FunctionComponent<ProfileEventsProps> = ({ child }) => {
   const hasEvents = (child: ChildByIdResponse) => {
-    return child.availableEvents || child.enrolments || child.pastEvents;
+    let childHasEvents = false;
+
+    if (child.availableEvents && child.availableEvents.edges.length !== 0) {
```
badrange: I think you can simplify this with something like 
```child.availableEvents.edges?.[0]```

File: src/domain/profile/events/__tests__/__snapshots__/ProfileEvents.test.tsx.snap [hunk 2]
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Renders snapshot correctly 1`] = `<ProfileNoEvent />`;
```
badrange: This doesn't actually render.. and that might be what we want in this case?

jpsalo: Could you elaborate? I am able to run the test on my machine.

PR #196, opened 2020-03-05
KK-282 | Add participation confirmation page dialog

badrange:
- Make the enrol button on event page do something
- Show actual remaining capacity on event page
- Simplify hasEvents logic code in ProfileEvents
- Fix typing issues in tests / add typing

### DIFF
--- a/src/domain/event/EventOccurrence.tsx
+++ b/src/domain/event/EventOccurrence.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { useTranslation } from 'react-i18next';
+import { Link } from 'react-router-dom';
 
 import { eventQuery_event_occurrences_edges_node as OccurrencesEdgeNode } from '../api/generatedTypes/eventQuery';
 import Button from '../../common/components/button/Button';
@@ -24,11 +25,13 @@ const EventOccurrence: React.FunctionComponent<EventOccurrenceProps> = ({
       <td className={styles.occurrenceDate}>{date}</td>
       <td className={styles.occurrenceTime}>{time}</td>
       <td className={styles.occurrenceVenue}>{occurrence.venue.name}</td>
-      <td className={styles.occurrenceFree}>999</td>
+      <td className={styles.occurrenceFree}>{occurrence.remainingCapacity}</td>
       <td className={styles.occurrenceSubmit}>
-        <Button type="submit" className={styles.submitButton}>
-          {t('event.register.occurrenceTableHeader.buttonText')}
-        </Button>
+        <Link to={`${occurrence.id}/enrol`}>
+          <Button type="submit" className={styles.submitButton}>
+            {t('event.register.occurrenceTableHeader.buttonText')}
+          </Button>
+        </Link>
       </td>
     </tr>
   );
--- a/src/domain/event/EventOccurrence.tsx
+++ b/src/domain/event/EventOccurrence.tsx
@@ -9,10 +9,12 @@ import styles from './eventOccurrence.module.scss';
 
 interface EventOccurrenceProps {
   occurrence: OccurrencesEdgeNode;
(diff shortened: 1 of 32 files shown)


src/domain/profile/events/ProfileEventsList.tsx
```diff
@@ -23,21 +24,23 @@ import Icon from '../../../common/components/icon/Icon';
...
   availableEvents: AvailableEventsTypes | null;
+  child: ChildByIdResponse;
   enrolments: EnrolmentsTypes;
   pastEvents: PastEventsTypes | null;
 }
 
 const EVENT_DURATION_MINUTES = 30; // TODO: huh?
 const ProfileEventsList: FunctionComponent<ProfileEventsListProps> = ({
   availableEvents,
+  child,
```
jpsalo: Do you think we can - or if we should - circumvent passing the whole child data for this component? It seems to only need the `id`, or, semantically, the url base.

badrange: I think we can! And I don't know if we should, I considered just passing the id and decided to pass the whole shebang thinking it could be convenient in the future. Would it help us prevent bugs or improve performance to only send the id?

jpsalo: I would say it is more likely to prevent bugs (and keep the components' responsibilities as small as possible) than to improve performance.

badrange: Makes sense. I'll update the PR with this in mind.

src/domain/event/enrol/enrol.module.scss
```diff
@@ -0,0 +1,62 @@
...
+  @include background($hel-light);
+  padding: $basePadding;
+  width: 100%;
+  display: grid;
+  grid-template-rows: 2;
+  grid-template-columns: 4;
+  & > * {
+    display: inline;
+  }
+}
```
jpsalo: Obsolete?

src/domain/event/EventOccurrence.tsx
```diff
@@ -24,11 +27,19 @@ const EventOccurrence: React.FunctionComponent<EventOccurrenceProps> = ({
...
+          {occurrence.remainingCapacity}
+        </td>
+      ) : (
+        <td></td>
+      )}
       <td className={styles.occurrenceSubmit}>
-        <Button type="submit" className={styles.submitButton}>
-          {t('event.register.occurrenceTableHeader.buttonText')}
-        </Button>
+        <Link to={`${occurrence.event.id}/occurrence/${occurrence.id}/enrol`}>
```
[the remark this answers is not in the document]
badrange: I just removed eventId, don't remember where it is from

src/domain/event/enrol/enrol.module.scss
badrange: Not exactly, but I refactored the OccurrenceInfo component to take a className as parameter

src/domain/event/EventOccurrenceList.tsx
```diff
@@ -26,9 +35,13 @@ const EventOccurrenceList: React.FunctionComponent<Occurrences> = eventOccurrenc
           </th>
         </tr>
```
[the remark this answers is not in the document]
badrange: Did I understand you?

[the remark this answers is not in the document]
jpsalo: Sorry, I meant the texts: 
1. `event.register.occurrenceTableHeader.eventInformation`
2. `event.register.occurrenceTableHeader.freePlaces`

src/domain/event/enrol/enrol.module.scss
jpsalo: Got it!

PR #198, opened 2020-03-10
KK-282 | Add support for enrolling and cancelling enrolment

badrange:
I added some TODOs with a reference to Jira-issues, hoping that we won't lose track of them

### DIFF
--- a/src/domain/event/EventIsEnrolled.tsx
+++ b/src/domain/event/EventIsEnrolled.tsx
@@ -2,7 +2,7 @@ import React, { FunctionComponent, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useHistory, useParams } from 'react-router-dom';
 import * as Sentry from '@sentry/browser';
-import { useQuery, useMutation } from '@apollo/react-hooks';
+import { useQuery } from '@apollo/react-hooks';
 
 import Icon from '../../common/components/icon/Icon';
 import styles from './event.module.scss';
@@ -13,7 +13,6 @@ import occurrenceQuery from './queries/occurrenceQuery';
 import { occurrenceQuery as OccurrenceQueryType } from '../api/generatedTypes/occurrenceQuery';
 import LoadingSpinner from '../../common/components/spinner/LoadingSpinner';
 import OccurrenceInfo from './partial/OccurrenceInfo';
-import unenrolOccurrenceMutation from './mutations/unenrolOccurrenceMutation';
 import UnenrolModal from './modal/UnenrolModal';
 
 const EventIsEnrolled: FunctionComponent = () => {
--- a/src/domain/event/EventIsEnrolled.tsx
+++ b/src/domain/event/EventIsEnrolled.tsx
@@ -1,0 +1,114 @@
+import React, { FunctionComponent, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { useHistory, useParams } from 'react-router-dom';
+import * as Sentry from '@sentry/browser';
+import { useQuery, useMutation } from '@apollo/react-hooks';
+
+import Icon from '../../common/components/icon/Icon';
+import styles from './event.module.scss';
+import PageWrapper from '../app/layout/PageWrapper';
+import backIcon from '../../assets/icons/svg/arrowLeft.svg';
+import Button from '../../common/components/button/Button';
+import occurrenceQuery from './queries/occurrenceQuery';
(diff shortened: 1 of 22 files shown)


src/domain/event/event.module.scss
```diff
@@ -110,3 +110,38 @@
...
+        margin-left: 0;
+      }
+    }
+  }
+  & > * {
+    margin-left: 0;
+    margin-right: $baseMargin;
+    white-space: nowrap;
+  }
+}
```
jpsalo: An alternative could be to create a small `<Label />` (icon + text) component since they are in freely ordered and in many place

src/domain/event/modal/UnenrolModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+            occurrenceId: occurrenceId,
+            childId: childId,
+          },
+        },
+      });
+      history.push('/profile');
+    } catch (error) {
+      console.error(error);
+    }
+  };
```
jpsalo: See comment for https://github.com/City-of-Helsinki/kukkuu-ui/pull/198/files#diff-c33dbf914bd403da31215a71525408f8R55-R69

src/domain/event/EventIsEnrolled.tsx
```diff
@@ -0,0 +1,114 @@
...
+            <OccurrenceInfo
+              className={styles.occurrenceInfo}
+              occurrence={data.occurrence}
+            />
+            <div className={styles.description}>
+              {data.occurrence.event.description}
+            </div>
+            <div className={styles.participantsPerInvite}>
+              {t(
+                `event.participantsPerInviteEnumLong.${data.occurrence.event.participantsPerInvite}`
```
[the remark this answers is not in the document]
badrange: What do you mean, is it missing from the translations sheet or?

src/domain/event/enrol/Enrol.tsx
```diff
@@ -41,7 +50,23 @@ const Enrol: FunctionComponent = () => {
...
+            occurrenceId: data?.occurrence?.id,
+            childId: params.childId,
+          },
+        },
+      });
+      history.push('/profile');
+    } catch (error) {
+      console.error(error);
+    }
+  };
```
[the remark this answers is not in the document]
badrange: Not sure. Would be nice to have something merged that can be tested out by the rest of the team - and to ensure that I don't hold you up..

src/domain/event/event.module.scss
badrange: Good idea. I propose that we do that in a follow up PR

src/domain/event/enrol/Enrol.tsx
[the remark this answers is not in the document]
badrange: I added some updates to the flow, and a TODO

src/domain/event/EventIsEnrolled.tsx
```diff
@@ -0,0 +1,114 @@
...
+              )}
+            </div>
+            <h2>{t('event.cancellation.heading')}</h2>
+            <Button
+              className={styles.cancelRegistration}
+              onClick={() => setIsOpen(true)}
+            >
+              {t('event.cancellation.buttonText')}
+            </Button>
```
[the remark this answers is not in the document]
badrange: I centered it
Not sure if it looks better now, though :-)

src/domain/event/enrol/Enrol.tsx
```diff
@@ -36,7 +38,17 @@ const Enrol: FunctionComponent = () => {
...
+      refetchQueries: [
+        { query: profileQuery },
+        {
+          query: childByIdQuery,
+          variables: {
+            id: params.childId,
+          },
+        },
+      ],
+    }
```
jpsalo: I'm unsure should `<Enrol />` be concerned about these.

src/domain/event/event.module.scss
```diff
@@ -86,7 +86,9 @@
 }
 
 .eventWrapper {
-  z-index: 1;
+  @include respond-above(md) {
+    z-index: 1;
```
jpsalo: :1st_place_medal:

src/domain/event/enrol/Enrol.tsx
badrange: I agree. But with our current implementation the UI will be out of date from the server without this.

badrange:
I agree that the refetchQueries are risky. We need to think about this.

badrange:
And maybe even do some refactoring!

PR #199, opened 2020-03-11
KK-359 | Allow whitespace in city field

badrange:


src/common/components/form/__tests__/validationUtils.tests.tsx
```diff
@@ -1,6 +1,5 @@
 import {
```
jpsalo: This file should probably be renamed to `validationUtils.test.tsx` (omit **s** from "tests")

PR #200, opened 2020-03-12
KK-302 | Add collapsible component to event page

jpsalo:
- [x] ~This is a draft until #198 is merged so that the component can be moved into enrolled event page (See [EventsIsEnrolled @ #198](https://github.com/City-of-Helsinki/kukkuu-ui/blob/feature/KK-282-participation-confirmation-page-dialog/src/domain/event/EventIsEnrolled.tsx)).~


src/common/components/collapsible/Collapsible.tsx
```diff
@@ -0,0 +1,35 @@
...
+  item: Item;
+}
+
+const Collapsible: FunctionComponent<CollapsibleProps> = ({ item }) => {
+  const [isOpen, toggleCollapsible] = React.useState(false);
+
+  return (
+    <>
+      <button
+        aria-expanded={isOpen}
```
badrange: Good to add aria-expanded! 
and I think we need to add an aria-label too  if we want screen reader users to be able to know what this button does?

badrange:
Also remember to ensure that the collapsible info shows up on the correct place: Right now it shows up on the event page, but it should be on the EventIsEnrolled component :-)

badrange:
I think there is room for more accessibility improvements! 
At least, let's use a header tag headers of the sections (H2 maybe?)

I found this article: 
https://inclusive-components.design/collapsible-sections/

PR #201, opened 2020-03-12
KK-279, KK-280 | Handle enrolment failed, Add filter by date and time

badrange:
I went a bit overboard with this PR and started on the occurrence filter. 

With the current implementation the filter dropdown only contains occurrences that match the current selection, so if you filter on for example tomorrow there is no way to switch to another day.

### DIFF
--- a/src/domain/api/generatedTypes/eventQuery.ts
+++ b/src/domain/api/generatedTypes/eventQuery.ts
@@ -79,4 +79,6 @@ export interface eventQuery {
 
 export interface eventQueryVariables {
   id: string;
+  date?: any | null;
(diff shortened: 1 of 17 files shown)


src/domain/api/generatedTypes/eventQuery.ts
```diff
@@ -79,4 +79,6 @@ export interface eventQuery {
 
 export interface eventQueryVariables {
   id: string;
+  date?: any | null;
+  time?: any | null;
```
Nizar-Rahme: A defined type instead of `any` would be better.

PR #204, opened 2020-03-16
KK-317 | Don't check city eligibility when editing a child

badrange:
This is a slight improvement over the current situation - we don't care what people type in the city field. We will store the postal code, though. 

Longer term solution involves either storing the city in the back end or looking the city up based on the postal code.


badrange:
> This one works. I'm however unsure, is our overall approach very good towards validating the city.

I'm open to suggestions on how to improve this! Since you're new to the project you are not yet blind to how things have been done..!

PR #206, opened 2020-03-19
KK-368 | Make all available dates and times available also after filtering

badrange:


src/domain/event/Event.tsx
```diff
@@ -40,6 +64,13 @@ const Event: FunctionComponent = () => {
   );
 
   const updateFilterValues = (filterValues: FilterValues) => {
+    // If date or time is missing we force it to be present and undefined to
+    // work around this apollo bug:
+    // https://github.com/apollographql/react-apollo/issues/2300
+    // Without it you would not be able to go from having a date or time
+    // filter to seeing all occurrences again.
+    filterValues.date = filterValues.date ? filterValues.date : undefined;
+    filterValues.time = filterValues.time ? filterValues.time : undefined;
```
t0mim: Some considerations:
Should we create new object here and not change the parameter value. Like,
```
const eventFilterValue = {
date: filterValues.date || undefined,
time: filterValues.time || undefined
};

File: src/domain/event/Event.tsx [hunk 2]
```diff
@@ -56,6 +87,51 @@ const Event: FunctionComponent = () => {
...
+              occurrence.node.time,
+              data.event?.duration || null
+            ),
+            key: occurrence.node.id,
+          }
+        : { key: '', label: '', value: '' };
+    })
+    .filter((v, i, a) => a.findIndex(t => t.value === v.value) === i)
+    .sort((a, b) => {
+      return a.label && b.label
```
t0mim: Do we need this first check? Also this nested conditional operand is bit hard to read.

PR #208, opened 2020-03-20
KK-372 | Add support for paragraphs in descriptions

badrange:


badrange:
What do you mean by that? To push this nl2br thing up a layer and do it for all the strings we're getting from the backend?

jpsalo:
Apologies (apol**l**ogies) for short reply. Yes, that was what I meant. Perhaps Apollo Link (middleware) is not the best solution but it would me nice to pre-process the texts before they end up into the components.

badrange:
If we do it in Apollo Link I think we need to use dangerouslySetInnerHTML - otherwise React is going to escape those <p> or <br /> tags we add in that layer. 

So I guess that solution is out of the question. Maybe one of the nl2br-react npm-modules could be more "generic" than my tiny hack..

PR #218, opened 2020-03-30
 KK-320 | Immutable child's birthdate on edit form

jpsalo:


badrange:
Great to have this refactored! 

I found two bugs: 
* From the profile page it is now possible to add a child not in Helsinki
* The birthdate displayed in the modal is 1 month later than the actual birthday. So if the child is born 13.1, the edit modal shows 13.2 as birthdate

PR #224, opened 2020-04-02
KK-353 | Upgrade all dependencies to the most recent version

badrange:
As a workaround for the name collision between our colors css module and apollo, I chose to change the default import path. This should prevent similar name collision errors to come up in the future. 

I also added a change that prevents a new warning from React - but I'm not sure if I'm doing it right.


.env
```diff
@@ -9,4 +9,4 @@ REACT_APP_VERSION=$npm_package_version
 TRANSLATION_LANGUAGES=en,fi,sv
 TRANSLATIONS_URL=https://docs.google.com/spreadsheets/d/18dpV7aA-t_6zMRUO5hP7DMfx0q4hdu-AlLpRjkB_0zI/gviz/
 TRANSLATION_PROJECT_NAME=kukkuu
-SASS_PATH=./src/assets/styles
+SASS_PATH=./src/assets
```
badrange, line 12: This works around the name collision issue with our colors css module.

src/domain/event/state/EventReducers.tsx
```diff
@@ -11,9 +11,9 @@ export default createReducer(defaultChildEventData, {
   [EVENT_ACTIONS.CLEAR_EVENT]: (state) => (state = defaultChildEventData),
   [EVENT_ACTIONS.SAVE_CHILDREN_EVENTS]: (state, action) => {
     const childrenEvents: ChildEvents[] = [];
-    action.payload.edges.forEach((childEdge: edges) => {
+    action.payload?.edges?.forEach((childEdge: edges) => {
```
badrange, line 14: This change exists in a separate PR, was added here by mistake 🤦🏻‍♂️

src/domain/home/Home.tsx
```diff
@@ -33,13 +34,14 @@ const Home: FunctionComponent = () => {
     }
   };
 
+  useEffect(() => {
```
badrange, line 37: I am not sure if this is the right approach to solve this: We get rid of the warning but could it be that we are just brushing the problem under the carpet?

src/domain/profile/Profile.tsx
```diff
@@ -19,13 +19,20 @@ import settingsIcon from '../../assets/icons/svg/gear.svg';
...
 import { clearEvent, saveChildrenEvents } from '../event/state/EventActions';
+import { defaultProfileData } from './state/ProfileReducers';
 
 const Profile: FunctionComponent = () => {
   const [isOpen, setIsOpen] = React.useState(false);
   const { loading, error, data } = useQuery<ProfileQueryType>(profileQuery);
   const { t } = useTranslation();
   const dispatch = useDispatch();
 
+  useEffect(() => {
```
badrange, line 30: I am not sure if this is the right approach to solve this: We get rid of the warning but could it be that we are just brushing the problem under the carpet?

package.json
```diff
@@ -9,56 +9,59 @@
...
-    "react-toastify": "^5.4.1",
-    "redux-oidc": "^3.1.5",
+    "react-scripts": "^3.4.1",
+    "react-toastify": "^5.5.0",
+    "redux-oidc": "^4.0.0-beta1",
     "redux-persist": "^6.0.0",
     "typescript": "3.8.3"
   },
+  "resolutions": {
+    "graphql": "^14.6.0"
```
badrange, line 63: This prevents having two versions of graphql, which causes an error in apollo

PR #226, opened 2020-04-08
KK-300, KK-394 | Improve event enrolment layouts and structure

badrange:
* Less margin on child info page on desktops
* Make the QR code look more similar to the event image
* Hide participants from child upcoming events info
* Center remaining capacity number on occurrence list screen
* Refactor OccurrenceInfo - less code duplication
* Add option to hide items from OccurrenceInfo
* Show confirm text on confirm screen
* Update translations

### DIFF
--- a/src/domain/event/partial/InfoItem.tsx
+++ b/src/domain/event/partial/InfoItem.tsx
@@ -5,20 +5,20 @@ import Icon from '../../../common/components/icon/Icon';
 
 type Props = {
   className?: string;
-  icon: string;
+  iconSrc: string;
   iconAlt?: string;
   label?: string | undefined;
 };
 
 const InfoItem: FunctionComponent<Props> = ({
   className,
-  icon,
+  iconSrc,
   iconAlt = '',
   label = '',
 }) => {
   return (
     <div className={className}>
-      <Icon alt={iconAlt} className={styles.labelIcon} src={icon} />
+      <Icon alt={iconAlt} className={styles.labelIcon} src={iconSrc} />
       <div>{label}</div>
     </div>
   );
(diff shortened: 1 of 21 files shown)


src/domain/event/partial/InfoItem.tsx
```diff
@@ -0,0 +1,27 @@
...
+type Props = {
+  className?: string;
+  icon: string;
+  iconAlt?: string;
+  label?: string | undefined;
+};
+
+const InfoItem: FunctionComponent<Props> = ({
+  className,
+  icon,
```
t0mim, line 15: Could this be named iconPath or iconSrc so that it would be clearer what type of data is needed.

src/domain/event/partial/OccurrenceInfo.tsx
```diff
@@ -7,63 +7,75 @@ import clockIcon from '../../../assets/icons/svg/clock.svg';
...
+      ),
+    },
+    {
+      id: 'venue',
+      icon: locationIcon,
+      label: occurrence.venue.name || '',
+    },
+  ];
+
+  console.log('classname');
```
t0mim, line 62: This could use template string.

badrange: This one should not have been part of the pr 😅

src/domain/event/partial/InfoItem.tsx
badrange: Good idea. We're using it like this in other parts of the project, but I'll replace it here for now.

badrange:
Before

image

badrange:
### After

image

image

PR #230, opened 2020-04-16
KK-342 | Improve SuccessToast placement and size on all screen sizes

badrange:
Bonus:

* Make all card components clickable
* Ensure that setTimeout is always cleared

### DIFF
--- a/src/domain/event/enrol/SuccessToast.tsx
+++ b/src/domain/event/enrol/SuccessToast.tsx
@@ -21,35 +21,35 @@ const SuccessToast: FunctionComponent = () => {
   }, [isJustEnrolled]);
 
   return (
-    <div className={styles.successToastWrapper}>
-      <CSSTransition
-        in={showJustEnrolled}
-        timeout={{ appear: 300, enter: 300, exit: 300 }}
-        classNames={{ ...styles }}
-        onEntered={() =>
-          setTimeout(() => {
-            setShowJustEnrolled(false);
-          }, 3000)
-        }
-        onExited={() => dispatch(justEnrolled())}
-        unmountOnExit
+    <CSSTransition
+      className={styles.successToast}
+      onClick={() => {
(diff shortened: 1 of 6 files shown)


src/domain/event/enrol/SuccessToast.tsx
```diff
@@ -21,35 +21,35 @@ const SuccessToast: FunctionComponent = () => {
...
-        unmountOnExit
+    <CSSTransition
+      className={styles.successToast}
+      onClick={() => {
+        setShowJustEnrolled(false);
+      }}
+      in={showJustEnrolled}
+      timeout={{ appear: 300, enter: 300, exit: 300 }}
+      classNames={{ ...styles }}
+      onEntered={() => {
```
badrange, line 32: @jorilindell is this the solution you were thinking about?

jorilindell: Is the function to clear timeout ever called? 

Instead you could do something like this.
1. At the beginning of the component define variable to store timeout
```
let timer: NodeJS.Timeout
```

2. Write useEffect hook to clear timer when component is unmounted
```
useEffect(() => {
  // Cleanup function to run on unmount
  return () => {
    clearTimeout(timer)
  }
})
```

3. Remove return function from onEntered
```
onEntered={() => {
  // If there is risk that onEntered is called multiple time you can also clear timer here, so 
  // setShowJustEnrolled(false) is called only once
  // clearTimeout(timer)
  timer = setTimeout(() => {
    setShowJustEnrolled(false);
  }, 3000)
}
```

badrange: This suggestion caused typing errors; TS complains that I was trying to use the variable before assigning it. Which gave me the idea that I can clear it inside the onExited call, I guess I can trust that the component always will call that.

PR #231, opened 2020-04-20
KK-409 | Past event page should not include occurrence info

badrange:
Bonus:

* Upgrade all packages because timing is good when we change graphql queries
* Small fix to timeout cleanup of Success toast


src/domain/event/Event.tsx
```diff
@@ -69,7 +75,8 @@ const Event: FunctionComponent = () => {
     .filter((c) => c.childId === params.childId)
     .pop()
     ?.eventIds.some((e) => e === params.eventId);
-  if (isRegistered) {
+  if (isRegistered && !past) {
+    console.log('isRegistered');
```
Jokinen, line 79: Left in on  purpose?

badrange: Thanks!! Good spot, a proof that it is important to have someone else looking over my shoulder!

PR #235, opened 2020-04-24
KK-268 | Inform user better when login fails due to device time being off

badrange:
Bonus: Add a <ErrorMessage /> component for less code duplication.

How to test: 

1. Use a non-local Tunnistamo
2. Set your device time to be minimum 5 minutes early or late
3. Log in

The strings for the error messages is still TODO, but those translations will be updated later.

### DIFF
--- a/src/common/translation/i18n/fi.json
+++ b/src/common/translation/i18n/fi.json
@@ -8,8 +8,8 @@
   "appName": "Kulttuurin kummilapset",
   "authentication": {
     "deviceTimeError": {
-      "message": "Usko tai et usko mutta et voi kirjaudua sisään koska laitteesi kello on yli 5 minuuttia väärässä. Säätää kelloa ja kokeile uudestaan. ",
-      "shortMessage": "Log in error. Check your device time and try again."
+      "message": "Usko tai älä usko mutta et voi kirjautua sisään koska laitteesi kello on yli 5 minuuttia väärässä. Säädä kelloa ja kokeile uudestaan.",
+      "shortMessage": "Virhe kirjautuessa. Tarkista laitteesi kello ja kokeile uudestaan."
     },
     "errorMessage": "Tapahtui virhe. Yritä uudestaan",
     "loadUserError": {
--- a/src/common/translation/i18n/fi.json
+++ b/src/common/translation/i18n/fi.json
@@ -7,6 +7,10 @@
   },
   "appName": "Kulttuurin kummilapset",
   "authentication": {
+    "deviceTimeError": {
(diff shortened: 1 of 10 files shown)


src/common/translation/i18n/fi.json
```diff
@@ -7,6 +7,10 @@
   },
   "appName": "Kulttuurin kummilapset",
   "authentication": {
+    "deviceTimeError": {
+      "message": "Usko tai et usko mutta et voi kirjaudua sisään koska laitteesi kello on yli 5 minuuttia väärässä. Säätää kelloa ja kokeile uudestaan. ",
```
Jokinen, line 11:
```suggestion
      "message": "Usko tai älä usko mutta et voi kirjautua sisään koska laitteesi kello on yli 5 minuuttia väärässä. Säädä kelloa ja kokeile uudestaan.",
```

src/common/translation/i18n/fi.json
```diff
@@ -7,6 +7,10 @@
...
+      "shortMessage": "Log in error. Check your device time and try again."
```
Jokinen, line 12:
```suggestion
      "shortMessage": "Virhe kirjautuessa. Tarkista laitteesi kello ja kokeile uudestaan."
```

src/domain/auth/OidcCallback.tsx
```diff
@@ -8,30 +8,52 @@ import * as Sentry from '@sentry/browser';
 
 import userManager from './userManager';
 import PageWrapper from '../app/layout/PageWrapper';
+import ErrorMessage from '../../common/components/error/Error';
 
 function OidcCallback(props: RouteChildrenProps) {
   const { t } = useTranslation();
 
+  const [callbackMessage, setCallbackMessage] = useState({
+    message: <p>{t('authentication.redirect.text')}</p>,
```
Jokinen, line 17: Here and later you are putting components into state. Wayback when React used to advise against this 🤔Not sure if times have changed since then:

http://web.archive.org/web/20150419023006/http://facebook.github.io/react/docs/interactivity-and-dynamic-uis.html#what-shouldnt-go-in-state

The refactor here would be to render the message you want on the return line, and save such values into state that you are able to determine what error should be rendered.

badrange: Thank you for this feedback! Good point! Tried to find the PR where this advice was removed from the react docs, even! 

I think this should be fine now four years after the recommendation was removed from the docs, especially given how the hooks work.

src/common/translation/i18n/fi.json
badrange: Thanks! I've updated the translation sheet, adding a new commit with this fix.

src/common/translation/i18n/fi.json
badrange: Thanks! I've updated the translation sheet, adding a new commit with this fix.

PR #236, opened 2020-04-28
KK-430 | Require relationship when adding child

badrange:


src/domain/child/modal/__tests__/__snapshots__/ChildFormModal.test.tsx.snap
```diff
@@ -20,7 +20,7 @@ exports[`renders snapshot correctly 1`] = `
           "lastName": "",
           "postalCode": "",
           "relationship": Object {
-            "type": undefined,
+            "type": null,
```
badrange, line 23: This issue took ages to figure out - I was expecting that the required message would pop up when I set the fields to required. 

BUUUUUT the Formik didn't set "touched" upon submit until I randomly changed this one from undefined to null. Huh.

badrange: Might have been "cleaner" to set it to `"" ` but the typing of that field won't allow it, and it does seem to work nicely with only `null` here

tuovinensanttu: Seems like typing comes all they from `globalTypes.ts` where all the fields are set to `SomeOtherType | null`. It's just kind of wierd that there is no typing error when value is `undefined`🤔

badrange: I think we didn't have the typing error because the data structure from RegistrationReducers was piped through Formik as initialValues before we sent it through graphql. 

But, the real reason is probably that we have cheated somewhere so the typing in the form is not consistent with the graphql schema..

PR #238, opened 2020-04-28
KK-292 | Show informative error when user denies request in Tunnistamo

badrange:


src/domain/auth/OidcCallback.tsx
```diff
@@ -35,6 +35,16 @@ function OidcCallback(props: RouteChildrenProps) {
         <ErrorMessage message={t('authentication.deviceTimeError.message')} />
       );
       shortMessage = t('authentication.deviceTimeError.shortMessage');
+    } else if (
+      // Handle error caused by end user choosing Deny in Tunnistamo's Permission Request:
+      // We could send an event to Analytics for this.
+      error.message ===
```
badrange, line 41: Wonder if there is a better way to detect this error, like an error code or error type...

Jokinen: That's a good question, but I don't have a clue.

Jokinen: That's a good question, but I don't have a clue.

badrange: image

This is what the debugger shows me. Going through the stack trace I end up in a minified JS file in redux-oidc (from the dist folder, stupidly distributed package) and then the next level in the rabbit hole is oidc-client. 

But now I just realised that the error message might be from Tunnistamo, so I think a string check is the only way (and a bit flaky..)

PR #241, opened 2020-05-05
KK-437 | Show project name for child list

badrange:
Bonus: Remove query that is not in use

### DIFF
--- a/src/domain/api/__tests__/client.test.ts
+++ b/src/domain/api/__tests__/client.test.ts
@@ -1,5 +1,5 @@
 import client from '../client';
-import { childrenQuery } from '../../child/queries/ChildQueries';
+import profileQuery from '../../profile/queries/ProfileQuery';
 
 jest.mock('../../auth/state/AuthenticationSelectors', () => ({
   apiTokenSelector: () => 'foo',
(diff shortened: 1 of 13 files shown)


src/domain/profile/ProfileUtil.ts
```diff
@@ -26,8 +28,17 @@ export const normalizeProfileChild = (
   };
 };
 
-export const getProjectsFromProfileQuery = (children: Children) => {
-  return sortedUniq(
-    children?.edges?.map((child) => child?.node?.project.year).sort()
-  );
+export const getProjectsFromProfileQuery = (children: Children): Project[] => {
+  const empty: Project = {
```
badrange, line 32: This is a way to prevent TS errors (since the generated types are all with `| undefined`). So many things I don't know how to do.

src/domain/api/__tests__/client.test.ts
```diff
@@ -21,11 +21,13 @@ describe('graphql client', () => {
 
     try {
       await client.query({
-        query: childrenQuery,
+        query: profileQuery,
```
badrange, line 24: Removing childrenQuery since it was only used in this test. Always nice to clean up the attic a bit.

PR #248, opened 2020-05-25
KK-446 | Let guardian edit their email address

badrange:


src/domain/home/form/partial/birthdateFormField.module.scss
```diff
@@ -23,6 +23,13 @@
     display: grid;
     grid-template-columns: 1fr 0.2fr 1fr 0.2fr 1fr;
     align-items: baseline;
+
+    /* Hide number input spinners */
```
badrange, line 27: This is unrelated, and was added by accident. It is desirable though.

Jokinen: I think you are still able to see the spinners in FF. Noticed by accident.

PR #256, opened 2020-05-27
KK-476 | Make it possible to change the email address while signing up

badrange:
This is draft because it depends on a backend change

### DIFF
--- a/src/domain/registration/form/RegistrationForm.tsx
+++ b/src/domain/registration/form/RegistrationForm.tsx
@@ -204,7 +204,6 @@ const RegistrationForm: FunctionComponent = () => {
                     placeholder={t(
                       'registration.form.guardian.email.input.placeholder'
                     )}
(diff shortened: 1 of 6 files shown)


src/domain/registration/form/RegistrationForm.tsx
```diff
@@ -192,13 +194,18 @@ const RegistrationForm: FunctionComponent = () => {
...
+                    id="guardian.email"
+                    name="guardian.email"
+                    required={true}
+                    validate={validateEmail}
+                    label={t('registration.form.guardian.email.input.label')}
+                    component={InputField}
+                    placeholder={t(
+                      'registration.form.guardian.email.input.placeholder'
+                    )}
+                    value={values.guardian.email}
```
Jokinen, line 207: Are you sure that `formik` doesn't handle the value field for you automatically? By default it at least attempts to apply handlers and value based on the name prop.

PR #262, opened 2020-06-05
KK-491 | Ensure that the childrenEventSelector is populated on auto-login

badrange:


src/domain/app/header/userDropdown/UserDropdown.tsx
```diff
@@ -22,9 +31,25 @@ const UserDropdown: FunctionComponent<UserDropdownProps> = ({
...
-  const profileData = useSelector(profileSelector);
+  const dispatch = useDispatch();
+
+  const { loading, error, data } = useQuery<ProfileQueryType>(profileQuery, {
+    skip: !isAuthenticated,
+  });
   const { trackEvent } = useMatomo();
 
+  useEffect(() => {
+    console.count('UserDropdown useEffect');
```
Jokinen, line 42: Left in by accident?

tuovinensanttu:
> Wouldn't know how to do it better myself. I don't know Kukkuu well enough to be able to judge in great detail either, but using `useEffect` to sync state is usually not optimal.

I'm not sure if this will work in your case or not, but If you want to try and avoid `useEffect` you could update state during rendering (as described in the [Docs](https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops)).

PR #273, opened 2020-07-13
KK-211 | Refactor from our custom form components to HDS

badrange:
This is work in progress, some todo-comments are put in. I *should* have taken one form at a time, but I suppose I got a bit overly ambitious. 

The end goal is that we don't have any custom components left, that we can rely on only HDS-provided components

TODO

- [x] Alignment on fields on home page form
- [x] Create wrapper for TextInput to avoid code duplication
- [x] src/domain/registration/form/RegistrationForm.tsx validationSchema doesn't work
- [x] RegistrationForm.tsx 2. adding a child, relationship from modal disappears
- [x] Dropdown shows "required" error message before you have "touched" it
- [x] EventEnrol: After selecting date, how to show all again? Is my implementation fine? 
- [ ] Switch from own icons to HDS-versions
- [ ] H1 on registration form has margin-bottom 0 - from where? does it matter?


src/domain/registration/form/RegistrationForm.tsx
```diff
@@ -194,65 +239,81 @@ const RegistrationForm: FunctionComponent = () => {
...
                       )}
-                      component={InputField}
                       placeholder={t(
                         'registration.form.guardian.lastName.input.placeholder'
                       )}
+                      invalid={
+                        getIn(touched, 'guardian.lastName') &&
+                        getIn(errors, 'guardian.lastName')
+                      }
+                      helperText={t(getIn(errors, 'guardian.lastName'))}
```
Jokinen, line 307: If you create a wrapper for TextInput, you would most likely be able to provide default error logic for the component. That way you don't have to repeat yourself as much when setting invalid and helperText.

I made a quick and dirty version of this for the feedback component. I used a hook to be able to share the error logic between different kind of form components:
https://github.com/City-of-Helsinki/kuva-feedback-component/blob/develop/src/common/components/formikWrappers/Input.tsx
https://github.com/City-of-Helsinki/kuva-feedback-component/blob/develop/src/common/hooks/useField.tsx

There's only a single form in the app, so it didn't need to be that flexible of an implementation.

badrange: Great suggestion! 

This is something that would probably come to mind if I was reviewing someone else's PR :-)

src/domain/event/EventPage.tsx
```diff
@@ -36,17 +36,20 @@ const EventPage = ({ event, children, success }: EventProps) => {
...
+            <Button
+              variant="secondary"
+              aria-label={t('common.backButton.label')}
+              className={styles.backButton}
+              onClick={() => history.goBack()}
+            >
+              <Icon
+                src={backIcon}
+                className={styles.backButtonIcon}
+                alt={t('common.backButton.label')}
```
Jokinen, line 49: Do you need alt here if your button has an aria-label (or vice versa)?

src/domain/event/EventPage.tsx
```diff
@@ -36,17 +36,20 @@ const EventPage = ({ event, children, success }: EventProps) => {
...
-          </Button>
+          <div className={styles.backButtonInnerWrapper}>
+            <Button
+              variant="secondary"
+              aria-label={t('common.backButton.label')}
+              className={styles.backButton}
+              onClick={() => history.goBack()}
+            >
+              <Icon
+                src={backIcon}
```
Jokinen, line 47: Did you consider using icons from HDS? https://hds.hel.fi/storybook/react/?path=/story/components-icons-ui--iconarrowleft

src/domain/event/enrol/Enrol.tsx
```diff
@@ -124,14 +124,11 @@ const Enrol: FunctionComponent = () => {
...
-          <Button className={styles.submitButton} onClick={() => enrol()}>
+          <Button onClick={() => enrol()}>
             {t('enrollment.confirmationPage.confirm.button')}
           </Button>
 
-          <Button
-            className={styles.backButton}
-            onClick={() => history.goBack()}
-          >
+          <Button onClick={() => history.goBack()} variant={'secondary'}>
```
Jokinen, line 131: Same as before: anti-prettier prop (`variant={'secondary'}` -> `variant="secondary"`)

src/domain/home/form/partial/birthdateFormField.module.scss
```diff
@@ -22,7 +24,7 @@
   .inputWrapper {
     display: grid;
     grid-template-columns: 1fr 0.2fr 1fr 0.2fr 1fr;
-    align-items: baseline;
+    //align-items: baseline;
```
Jokinen, line 27: Comment that can likely be removed.

src/domain/profile/Profile.tsx
```diff
@@ -49,14 +48,15 @@ const Profile = () => {
...
-              <Button
-                className={styles.editProfile}
-                ariaLabel={t('profile.edit.button.text')}
-                onClick={() => setIsOpen(true)}
-              >
-                <span>{t('profile.edit.button.text')}</span>
-                <Icon src={settingsIcon} alt="" />
-              </Button>
+              <div className={styles.editProfile}>
+                <button
```
Jokinen, line 52: Did you mean to use the basic HTML button?

src/domain/profile/children/ProfileChildrenList.tsx
```diff
@@ -88,15 +86,12 @@ const ProfileChildrenList: FunctionComponent = () => {
         )}
       </div>
       <Button
+        variant={'secondary'}
```
Jokinen, line 89: Prettier formatting.

src/domain/profile/children/ProfileChildrenList.tsx
```diff
@@ -88,15 +86,12 @@ const ProfileChildrenList: FunctionComponent = () => {
...
         className={styles.addChild}
+        iconLeft={<IconPlusCircle />}
         onClick={() => setIsOpen(true)}
       >
-        <Icon
-          src={addIcon}
-          alt={t('child.form.modal.add.label')}
-          className={styles.addChildIcon}
-        />
         <span>{t('child.form.modal.add.label')}</span>
```
Jokinen, line 95: Do you still need this span?

src/domain/profile/children/child/ProfileChildDetail.tsx
```diff
@@ -75,17 +75,19 @@ const ProfileChildDetail: FunctionComponent = () => {
...
-          className={styles.backButton}
-          onClick={() => history.goBack()}
-        >
-          <Icon
-            src={backIcon}
-            className={styles.backButtonIcon}
-            alt={t('common.backButton.label')}
-          />
-        </Button>
+        <div className={styles.backButton}>
```
Jokinen, line 78: Button class on a div.

src/domain/profile/children/child/ProfileChildDetail.tsx
```diff
@@ -75,17 +75,19 @@ const ProfileChildDetail: FunctionComponent = () => {
...
+          <Button
+            variant={'secondary'}
+            aria-label={t('common.backButton.label')}
+            onClick={() => history.goBack()}
+          >
+            <Icon
+              src={backIcon}
+              className={styles.backButtonIcon}
+              alt={t('common.backButton.label')}
```
Jokinen, line 87: Two labels for this icon.

src/domain/profile/children/child/ProfileChildDetail.tsx
```diff
@@ -102,18 +104,23 @@ const ProfileChildDetail: FunctionComponent = () => {
...
+                  <Button
+                    className={styles.editChildInfoButton}
+                    variant={'secondary'}
+                    aria-label={t('profile.edit.button.text')}
+                    onClick={() => setIsOpen(true)}
+                    iconRight={
+                      <Icon
+                        src={settingIcon}
+                        className={styles.settingIcon}
+                        alt={t('profile.edit.button.text')}
```
Jokinen, line 117: Two labels for this icon.

src/domain/registration/form/RegistrationForm.tsx
```diff
@@ -33,14 +30,49 @@ import { profileQuery as ProfileQueryType } from '../../api/generatedTypes/profi
...
+      .required('validation.general.required')
+      .max(255, 'validation.maxLength'),
+    language: yup.string().max(255, 'validation.maxLength'),
+  }),
+  children: yup.array().of(
+    yup.object().shape({
+      postalCode: yup.string().required('validation.general.required'),
+      // relationship: yup.object().shape({
+      //   type: yup.string().required('validation.general.required').nullable(),
+      // }),
```
Jokinen, line 63: Commented code that can likely be removed.

src/domain/registration/form/RegistrationForm.tsx
```diff
@@ -80,12 +112,16 @@ const RegistrationForm: FunctionComponent = () => {
         <div className={styles.registrationForm}>
           <Formik
             initialValues={initialValues}
+            // TODO: Fix validationSchema
+            // validationSchema={schema}
             validate={() => {
               if (!isFilling) {
                 setFormIsFilling(true);
               }
+              console.log('validating');
```
Jokinen, line 121: Console log.

src/domain/registration/form/RegistrationForm.tsx
```diff
@@ -80,12 +112,16 @@ const RegistrationForm: FunctionComponent = () => {
...
             }}
-            onSubmit={(values) => {
+            onSubmit={(values: RegistrationFormValues) => {
+              console.log('submitting', values);
```
Jokinen, line 124: Console.log

src/domain/registration/form/RegistrationForm.tsx
```diff
@@ -33,14 +30,49 @@ import { profileQuery as ProfileQueryType } from '../../api/generatedTypes/profi
...
+  agree: yup.boolean().oneOf([true], 'validation.required'),
+});
 
 const RegistrationForm: FunctionComponent = () => {
   const { i18n, t } = useTranslation();
   const currentLocale = getCurrentLanguage(i18n);
   const history = useHistory();
   const dispatch = useDispatch();
   const [isOpen, setIsOpen] = useState(false);
+  const [checked, setChecked] = useState(false);
```
Jokinen, line 75: Seems like you are planning to have this value in the validation schema which alludes that you are planning to have the value in formik's state. In that case you can use formik's state to store the value and don't need this state. Note that you likely want to clean this field out before forwarding the values.

You can use Formik and yup in a few other ways to handle a value you don't want to forward, but want to validate. Cleaning it out before sending is probably the easiest and simplest one.

src/domain/registration/form/partial/ChildFormField.tsx
```diff
@@ -1,46 +1,57 @@
...
+type ChildFormFieldProps = {
   child: Child;
   childIndex: number;
   arrayHelpers: ArrayHelpers;
-}
+} & Pick<FormikState<RegistrationFormValues>, 'errors' | 'touched'> &
+  Pick<
+    FormikHelpers<RegistrationFormValues>,
+    'setFieldTouched' | 'setFieldValue'
+  >;
```
Jokinen, line 28: Could this API become clearer if you just accepted the formik objects as is?

Etc.
```
formikState: FormikState<RegistrationFormValues>,
formikHelpers: FormikHelpers<RegistrationFormValues>,
```

src/domain/registration/form/partial/ChildFormField.tsx
```diff
@@ -57,54 +68,79 @@ const ChildFormField: React.FunctionComponent<ChildFormFieldProps> = ({
...
+            getIn(errors, `children[${childIndex}].postalCode`)
+          }
+          helperText={
+            getIn(touched, `children[${childIndex}].postalCode`) &&
+            t(getIn(errors, `children[${childIndex}].postalCode`))
+          }
         />
 
-        <EnhancedInputField
+        <Field
```
Jokinen, line 118: An unnecessary `<Field />`.

src/domain/registration/form/partial/ChildFormField.tsx
```diff
@@ -57,54 +68,79 @@ const ChildFormField: React.FunctionComponent<ChildFormFieldProps> = ({
...
           id={`children[${childIndex}].relationship.type`}
           name={`children[${childIndex}].relationship.type`}
           label={t('registration.form.child.relationship.input.label')}
           required={true}
-          component={SelectField}
           options={getTranslatedRelationshipOptions(t)}
-          placeholder={t(
-            'registration.form.child.relationship.input.placeholder'
-          )}
+          onBlur={() =>
```
Jokinen, line 126: The `<Field />` component you use inside `FormikDropdown` has about the same implementation as you have here. I know you have included it because `onBlur` didn't work, but don't forget to clean these out.

src/common/components/button/Button.tsx
```diff
@@ -10,6 +10,7 @@ interface ButtonProps extends ReactButton {
   children: ReactNode;
   ariaLabel?: string;
 }
+// TODO: Delete me when HDS buttons can be added without superficial whitespace
```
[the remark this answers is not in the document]
badrange: The HDS buttons right now include some padding and margin that you can't "turn off". That's why I have added some margins with calc (x * -1) etc. Might be fine to not use HDS buttons for the remaining places where Button is used and just rename this component to something like NavigationButton

src/common/components/card/card.module.scss
```diff
@@ -90,16 +91,11 @@ $imageMaxSize: 300px;
...
-  button {
-    white-space: nowrap;
-    min-width: fit-content;
-    padding: 0 2rem;
-    @include respond-below(sm) {
-      width: 100%;
-    }
+  min-width: fit-content;
+  padding: 0 2rem;
+  @include respond-below(sm) {
```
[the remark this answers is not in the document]
badrange: The developer who started this project wasn't using mobile first, more like desktop first. I've been gradually refactoring over to mobile first (some examples in this pr), but not as the highest priority. I'll take a look at this one.

src/common/components/form/formikHelper.ts
```diff
@@ -0,0 +1,32 @@
+import { ReactNode } from 'react';
+import { FormikProps } from 'formik';
+import lodash from 'lodash';
```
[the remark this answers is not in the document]
badrange: Ooooh good catch!!

src/domain/child/form/ChildForm.tsx
```diff
@@ -206,13 +257,13 @@ const ChildForm: FunctionComponent<ChildFormProps> = ({
           </div>
 
           {isEditForm && (
-            <Button
+            <button
```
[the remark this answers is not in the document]
badrange: Well, it does add that extra whitespace, but yes it is good to be consistent

src/domain/event/EventIsEnrolled.tsx
```diff
@@ -56,12 +56,11 @@ const EventIsEnrolled = () => {
         )}
       </div>
       <h2>{t('event.cancellation.heading')}</h2>
-      <Button
-        className={styles.cancelRegistration}
-        onClick={() => setIsOpen(true)}
-      >
-        {t('event.cancellation.buttonText')}
-      </Button>
+      <div className={styles.cancelRegistration}>
```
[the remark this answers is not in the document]
badrange: It functions as a wrapper class, would it be better if I renamed it to cancelRegistrationButtonWrapper or something?

src/domain/event/EventOccurrence.tsx
```diff
@@ -36,13 +36,16 @@ const EventOccurrence: React.FunctionComponent<EventOccurrenceProps> = ({
...
+            className={styles.linkButton}
+            to={`${occurrence.event.id}/occurrence/${occurrence.id}/enrol`}
+          >
             <Button type="submit" className={styles.submitButton}>
               {t('event.register.occurrenceTableHeader.buttonText')}
             </Button>
           </Link>
         ) : (
-          <Button className={styles.fullButton} disabled>
+          <Button className={styles.fullButton} disabled={true}>
```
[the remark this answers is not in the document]
badrange: No message shows up, could be our project needs a bit stricter config? Fixed it anyway!

src/domain/event/EventOccurrence.tsx
```diff
@@ -36,13 +36,16 @@ const EventOccurrence: React.FunctionComponent<EventOccurrenceProps> = ({
         }
 
         {hasCapacity ? (
-          <Link to={`${occurrence.event.id}/occurrence/${occurrence.id}/enrol`}>
+          <Link
+            className={styles.linkButton}
```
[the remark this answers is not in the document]
badrange: Indeed. It already looked like a button, so that part hasn't changed in this PR.

src/domain/event/EventPage.tsx
badrange: I changed a couple of icons and then decided to wait with the rest until this PR is merged, it has grown so large. So a follow up PR could be for changing the icons.

src/domain/profile/Profile.tsx
badrange: Yes. I'll add a comment as to why, it is because HDS adds some whitespace that breaks the layout and is hard/impossible to remove AFAIK

src/domain/registration/form/RegistrationForm.tsx
badrange: Or fixed to actually work ..

src/common/components/form/formikHelper.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+const defaultErrorRender = (error: string) => error;
+
+export function getError<FormValues>(
+  formikProps: FormikProps<FormValues>,
+  fieldName: string,
+  render: (error: string) => ReactNode = defaultErrorRender
+) {
+  const errorMessage = lodash.get(formikProps.errors, fieldName);
```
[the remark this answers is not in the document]
badrange: Good idea. I just noticed that these helpers are not in use anywhere at the moment.

src/assets/styles/tokens.scss
```diff
@@ -0,0 +1,36 @@
...
+  --color-label-button-primary-focus: var(--color-black-90);
+  --color-label-button-primary-hover-focus: var(--color-black-90);
+  --color-label-button-primary-disabled: var(--color-black-90);
+  --color-label-button-secondary-default: var(--color-black-90);
+  --color-label-button-secondary-hover: var(--color-black-90);
+  --color-label-button-secondary-focus: var(--color-black-90);
+  --color-label-button-secondary-hover-focus: var(--color-black-90);
+  --color-label-button-secondary-disabled: #999898;
+  --color-background-checkbox-selected: var(--color-summer);
+  --color-background-checkbox-hover: var(--color-summer);
```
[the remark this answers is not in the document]
badrange: No it will not work. I wonder what is the best approach to fix this for the future, is something like this ok: 

```      <Button
        variant="secondary"
        aria-label={t('child.form.modal.add.label')}
        className={styles.addChild}
        style={kukkuuStyles}
        iconLeft={<IconPlusCircle />}
        onClick={() => setIsOpen(true)}
      >
```

or do you have a better suggestion?

src/common/components/button/Button.tsx
```diff
@@ -1,28 +1,69 @@
-import React, { FunctionComponent, ReactNode } from 'react';
-import joinClassNames from 'classnames';
+import React from 'react';
+import { Button as HdsButton, ButtonProps } from 'hds-react';
 
-import styles from './button.module.scss';
+const kukkuuSecondaryButtonStyles = {
+  '--background-color': 'var(--color-white)',
```
Jokinen, line 5: I didn't know this syntax was possible. Looks good to me, but I can't give a good opinion about it.

src/common/components/card/Card.tsx
```diff
@@ -57,10 +57,14 @@ const Card: FunctionComponent<CardProps> = ({
       </div>
 
       <div className={styles.end}>
-        <button aria-label={actionText} className={styles.actionWrapper}>
+        <Button
+          variant="supplementary"
+          aria-label={actionText}
```
Jokinen, line 62: The same text is used inside this component and in the aria label. Won't this result in the same string being read twice?

src/common/components/dropdown/dropdown.module.scss
```diff
@@ -34,26 +18,35 @@ $navbarIconSize: 1rem;
...
+    }
 
     button {
       width: 100%;
       min-height: $inputHeight;
       padding: $buttonPadding;
       display: flex;
       span {
-        flex-basis: calc(100% - 1rem);
+        // flex-basis: calc(100% - 1rem);
```
Jokinen, line 42: Rogue comment.

src/common/components/formikWrappers/DropdownWithFormik.tsx
```diff
@@ -0,0 +1,59 @@
...
+  name: string;
+  id: string;
+  onChange: (e: SimulatedEvent) => void;
+  value: string;
+  helperText?: string | ReactNode;
+  labelText?: string | ReactNode;
+  options: Array<{ label: string; value: string }>;
+};
+
+type Props = Omit<DropdownProps, 'value' | 'onChange'> & {
```
Jokinen, line 27: Personally I wouldn't make this limitation. Sometimes you may need custom `onChange` logic, or you need to provide a default value for `value` for instance.

This might have been useful when you refactored though, so I understand if it's included for that.

src/domain/app/hdsTheme/HdsTheme.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+//import Dropdown from "../../common/components/dropdown/Dropdown";
+//import { FormTheme } from "../feedbackForm/types";
```
Jokinen: Comments

src/domain/app/hdsTheme/HdsTheme.tsx
```diff
@@ -0,0 +1,24 @@
...
+} as React.CSSProperties;
+
+/*
+--background-selected:var(--color-success);
+--background-hover:var(--color-success-dark); 
+--border-color-selected:var(--color-success);
+--border-color-selected-hover:var(--color-success-dark);
+--border-color-selected-focus:var(--color-success);
+--focus-outline-color:var(--color-black-20);
+*/
```
Jokinen: Comments.

src/domain/profile/Profile.tsx
```diff
@@ -48,16 +49,15 @@ const Profile = () => {
...
-                  onClick={() => setIsOpen(true)}
-                >
-                  <span>{t('profile.edit.button.text')}</span>
-                  <Icon src={settingsIcon} alt="" />
-                </button>
-              </div>
+              <Button
+                variant="supplementary"
+                className={styles.editProfileButton}
+                aria-label={t('profile.edit.button.text')}
```
Jokinen, line 55: This label is probably not necessary?

src/domain/profile/children/child/ProfileChildDetail.tsx
```diff
@@ -104,23 +104,16 @@ const ProfileChildDetail: FunctionComponent = () => {
...
-                    }
-                  >
-                    {t('profile.edit.button.text')}
-                  </Button>
-                </div>
+
+                <Button
+                  variant="supplementary"
+                  className={styles.editChildInfoButton}
+                  aria-label={t('profile.edit.button.text')}
```
Jokinen, line 111: Here's another one of those labels.

Jokinen (approved):
Only minor comments.

src/common/components/card/Card.tsx
badrange: Good point! Leftover from old sins.

src/domain/app/hdsTheme/HdsTheme.tsx
badrange: Yeah this file was added by mistake :-)

src/assets/styles/tokens.scss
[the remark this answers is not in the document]
badrange: Came up with a better solution.

src/common/components/formikWrappers/FormikDropdown.tsx
```diff
@@ -17,23 +42,38 @@ export type HdsOptionType = {
...
+  labelText,
+  options,
+  ...rest
+}: Props) {
+  const { t } = useTranslation();
+  const [field, meta, helpers] = useField(name);
+  const [isInvalid, setIsInvalid] = useState(false); // Lol where do we call this
+
+  const handleChange = (selectedItem: any) => {
+    console.log(selectedItem);
```
Jokinen, line 60: console log

src/common/components/formikWrappers/FormikDropdown.tsx
```diff
@@ -17,23 +42,38 @@ export type HdsOptionType = {
...
+  onChange = () => ({}),
+  value,
+  helperText,
+  labelText,
+  options,
+  ...rest
+}: Props) {
+  const { t } = useTranslation();
+  const [field, meta, helpers] = useField(name);
+  const [isInvalid, setIsInvalid] = useState(false); // Lol where do we call this
```
Jokinen, line 57: The form is validated against the yup schema or the validate function. I don't think that you need to call any error logic on this level by default. The `meta` object actually contains information about the error state.

https://formik.org/docs/api/useField#fieldmetapropsvalue

badrange: That makes sense. I did it like this for two reasons, first due to typing errors - second as a first step in trying to fake onBlur / visited for these dropdowns.

Jokinen: For doing those things, `helpers` may be useful. That object contains for instance `setTouched` which is what you'd likely end up calling if you "faked" onBlur.

src/common/components/form/fields/dropdown/FormikDropdown.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export type OptionType = {
+  value: string;
+  label: string;
+};
+
+export type HdsOptionType = {
+  //eslint-disable-next-line
+  [key: string]: any;
+};
```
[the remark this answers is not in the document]
badrange: This one is now removed!

src/domain/registration/form/RegistrationForm.tsx
badrange: This wrapper is now done.

src/domain/child/form/ChildForm.tsx
```diff
@@ -18,6 +26,14 @@ import {
 import { formatTime, newMoment } from '../../../common/time/utils';
 import { BACKEND_DATE_FORMAT } from '../../../common/time/TimeConstants';
 
+const schema = yup.object().shape({
+  homeCity: yup.string().required('validation.general.required'),
+  postalCode: yup.string().required('validation.general.required'),
+  // relationship: yup.object().shape({
```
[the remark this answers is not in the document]
badrange: It's now fixed. Good spot.

src/domain/child/form/ChildForm.tsx
```diff
@@ -96,13 +114,24 @@ const ChildForm: FunctionComponent<ChildFormProps> = ({
     return isEditForm && immutableFields.includes(fieldName);
   };
 
+  const relationshipOptions = getTranslatedRelationshipOptions(t);
+
   return (
     <Formik
       validate={validateForm}
       initialValues={initialValues}
+      validationSchema={schema}
```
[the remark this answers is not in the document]
badrange: The validate function does something more than the schema, it handles the strange birthdate fields. For now I think we can live with the current implementation - would be nice to refactor it to only use schema, maybe at a later point in time..

src/domain/event/EventEnrol.tsx
```diff
@@ -59,28 +60,40 @@ const EventEnrol = ({
...
                     className={styles.dateField}
                     id="date"
                     name="date"
-                    placeholder="placeholder"
-                    label="Choose date"
-                    options={options.dates}
-                    component={SelectField}
-                    value={values.date}
+                    placeholder={t('common.select.default.text')}
+                    onChange={(option: HdsOptionType) =>
```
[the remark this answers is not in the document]
badrange: Yes! Good spot!

src/domain/event/EventEnrol.tsx
```diff
@@ -59,28 +60,40 @@ const EventEnrol = ({
               handleSubmit(values);
             }}
           >
-            {({ handleSubmit, values }) => {
+            {({ handleSubmit, setFieldValue, values }) => {
               return (
                 <form onSubmit={handleSubmit} id="eventPageForm">
-                  <EnhancedInputField
+                  <Field
```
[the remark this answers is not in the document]
badrange: Ah, good point.

src/common/components/form/fields/dropdown/FormikDropdown.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+function FormikDropdown(props: Props) {
+  // HDS Dropdown expects default value to be an object. Find correct option object from array.
+  const getSelectDefault = (options: OptionType[], value?: string) => {
+    return options.find((option: OptionType) => option.value === value);
+  };
+
+  return (
+    <Field name={props.name}>
```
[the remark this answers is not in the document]
badrange: This was a good idea, it made the whole component much simpler. Thank you.

src/domain/event/EventPage.tsx
badrange: No, removed.

src/domain/home/form/partial/birthdateFormField.module.scss
badrange: Indeed.

src/domain/profile/children/ProfileChildrenList.tsx
badrange: Nice find. Was able to remove some css imports and variables!

src/domain/profile/children/child/ProfileChildDetail.tsx
badrange: Changed classname to a wrapper.

src/common/components/formikWrappers/DropdownWithFormik.tsx
badrange: This file is now removed.

src/domain/registration/form/partial/ChildFormField.tsx
badrange: Probably. But that requires refactoring of the registration form, I would like to save that for later.

Jokinen:
Looks good so far 👍

src/common/components/formikWrappers/FormikDropdown.tsx
```diff
@@ -53,17 +53,23 @@ function FormikDropdown({
...
-      options={options}
-      onChange={handleChange}
-      invalid={Boolean(meta.error)}
-      helper={Boolean(meta.error) && t(meta.error || '')}
-      multiselect={false}
-      {...rest}
-    />
+    <div
+      onBlur={() => {
+        helpers.setTouched(true);
```
Jokinen, line 58: This is pretty interesting to note actually.

https://reactjs.org/docs/events.html#focus-events

In vanilla JS, a div element would not have the ability to `blur`, but in React all elements do.

In your testing, how does this approach behave? Will `onBlur` be called whenever any element within this div loses focus? In that case the `Dropdown` would become touched after an item has been selected instead of after the focus leaves the component entirely.

They are also doing some changes to event handling in the upcoming versions of React
https://reactjs.org/blog/2020/08/10/react-v17-rc.html#aligning-with-browsers

They explicitly promise:
>Although React 17 switched from focus to focusin under the hood for the onFocus event, note that this has not affected the bubbling behavior. In React, onFocus event has always bubbled, and it continues to do so in React 17 because generally it is a more useful default. See this sandbox for the different checks you can add for different particular use cases.

I think it's implied that the same applies to `onBlur`, at least the sandbox has an example. So it's unlikely that they will remove this special behaviour.

src/domain/registration/form/RegistrationForm.tsx
badrange: I refactored this to using a special Terms checkbox component, based on your implementation in another project, pretty clever and nice solution.

src/domain/registration/form/partial/ChildFormField.tsx
badrange: So resolving it

src/common/components/formikWrappers/FormikDropdown.tsx
badrange: Found a much simpler way to handle the "faked" onBlur!

src/common/components/formikWrappers/FormikDropdown.tsx
badrange: Good observation. I noticed now that the Invalid state is turned on when you activate the dropdown, and if you choose "select" and then select a value, it stays as invalid until you "tab" out of the dropdown. I'll see if I can improve the logic for invalid.

badrange: I updated the PR with logic from that link, thanks a lot!

PR #277, opened 2020-08-20
KK-527 | Use Finnish time format

Jokinen:
Moment uses the `en` locale by default. I changed the bootstrap code so that moment's locale and the language of the application are kept in sync.

As a small encapsulation measure I added a `setLocale` utility function which can possibly be used to configure other localisation dependent libraries (like numbers).

The file structure is maybe a bit odd here. I don't have a good intuitive feel for how internationalisation and localisation differ so I don't know which is the correct terminology here. But I did think that it didn't belong under `translations`.


badrange (approved):
Nice!

PR #284, opened 2020-09-04
KK-541 | Event occurrence time is shown incorrectly

tuovinensanttu:
- Wen't through every format executed with moment and replaced strings with constants.
- Changed `DEFAULT_DATE_FORMAT ` not show leading zeroes.


Jokinen (approved):
Sorry, I missed this!

It's looking good!

Personally separating time with `: ` would be my first choice, but I doubt that using a `.` is any less correct!

PR #285, opened 2020-09-04
KK-540 | Redirect to error view when accessing unknown child

Jokinen:
The user can login to the service using other services like Google or
Facebook. Once they have logged in using a service, they must always
use the same one if they want to access the "kukkuu" account that is
created then.

Sometimes users login with a service they did not use during their
initial sign up process. In this scenario, they are not able to access
their original data.

A usecase that's especially problematic is when the user receives a
link into their mail they can use to sign up their child into an event.
If they login by using the wrong service, they end up in a view that
doesn't work and gives little in the way of clues as to why.

After this commit, the user would be redirected into the wrong login
method page, which guides them to sign in again.

----

I've tested these things by hand:
* Links like this lead to a redirect to the error page: http://localhost:3000/fi/profile/child/ihan-mita-vaan/event/ihan-mita-vaan
* From my own profile, I am able to access my children's profile page
* Any link like this will lead to a redirect to the error page: http://localhost:3000/fi/profile/child/imagined-id
* Logging out works from the main navigation
* Logging out works from the error page

I've not written tests. This behaviour should likely be tested with `e2e` because it relies on GraphQL, but TestCafe has not been integrated yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+# Unreleased
+
+### Added
+
+- When users access a child's information that is not attached to their account, they are redirected into a view that asks them to authenticate again
+
 # 1.2.1
 
 ### Fixed
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "graphql": "^14.5.8",
     "graphql-tag": "^2.10.4",
     "hds-core": "^0.12.1",
+    "hds-design-tokens": "^0.13.0",
     "hds-react": "^0.12.1",
     "helsinki-utils": "City-of-Helsinki/helsinki-utils-js.git#0.1.0",
     "i18next": "^19.7.0",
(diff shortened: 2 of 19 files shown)


tuovinensanttu (approved):
Looking real nice 👍

badrange:
Can this actually happen in production - if you try to sign in with a different auth method Tunnistamo will stop you with an error message.. ?

Jokinen:
I don't quite understand what the scenario is you are thinking of.

For instance, how would Tunnistamo know that you have used the wrong authentication method if you have logged out of Tunnistamo?

badrange:
If you log in to tunnistamo with pekka@gmail.com as email address through YLE Tunnus, and the next time attempt to log in with Facebook or Google, Tunnistamo will stop you and say that you've already used this email with YLE tunnus.

badrange:
BUT If you use a different email, then there is another story. Just don't understand how you'll find a link to a "wrong" profile in another email. 

In any case, good to get this fixed, because it is absolutely a possible scenario that a user will use an "old" link and a different email address

tuomas777:
After clicking the link you first need to sign in. If you use a different auth method (with a different email address) than the one you used when registering your child (perhaps more than half a year ago), you get hit by this.

badrange:
Right, now I get it, was a bit slow here!

PR #290, opened 2020-09-11
KK-550 | Fix mobile layout for event enrollment

Jokinen:
- Event page should no longer overflow on mobile. In other words, the page should not have a horizontal scroll bar on mobile.
- Buttons and event information should now be extended to page margins on mobile


Jokinen:
Rebased to enable CI.

tuovinensanttu (approved):

PR #291, opened 2020-09-11
KK-546 | Fix vague naming of profile menu for screen readers

Jokinen:
Previously the profile menu relied on the `aria-label` attribute in order to communicate its state. Moreover, the naming applied to it was relatively vague. It was simply named as "menu".

In these changes I've adjusted the use of aria-attributes. I've added `aria-haspopup` and removed the use of state within the `aria-label`. The `aria-expanded` attribute already communicates this. We've also adjusted the button naming. Instead of "menu" the accessible content is now read as _label of button_ + _label of icon_. In the case of the user menu, for instance, 

>George profile menu

I checked the mobile menu as well while doing these changes. Unfortunately it ended up being less simple than the regular navigation changes. I think I need more context before I am able to make a proper fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 ### Fixed
 
 - Registration form final approval text
+- Provide better accessible name for profile menu
 
 # 1.3.0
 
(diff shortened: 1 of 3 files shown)


tuovinensanttu (approved):

Jokinen:

PR #294, opened 2020-09-15
KK-549 | Fix empty child list on mobile

Jokinen:
Previously the logged in user's profile was synced into the redux state
when the details for their user menu were fetched. On mobile, this menu
is not rendered by default, so their details were not ever synced and
they ended up seeing an empty list of children instead.

It's a bit odd that the user details are stored in two places: (1) in
Apollo's state and in (2) Kukkuu's redux state. I think that ideally
this duplication would not exist. But because I don't understand enough
of the context of the application yet, I didn't remove this
duplication entirely.

As a fix I created the useProfile hook. It fetches the profile and
stores it into the redux state--roughly the logic that was contained
within the user menu before. After this change, the profile is loaded
whenever it is accessed so there's no chance that a view will request
the profile before it has been requested. Further, this structure will
make the consumers of profile data more agnostic about its source which
will make it easier to refactor state management later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - Registration form final approval text
 - Provide better accessible name for profile menu
 - Mobile layout for event enrollment
+- Profile view missing children on mobile
 
 # 1.3.0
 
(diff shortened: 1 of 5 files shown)


badrange:
Absolutely incredible that we didn't notice that bug before? And what an error in thinking! 

You are absolutely right in thinking that it would be better to only rely on apollo for this data. If my memory serves me we chose to use redux because either apollo wasn't mature enough at the time or because we just didn't know how to use it..

badrange:
I like your approach, good job at figuring out how to improve on this functionality!

tuovinensanttu (approved):
Good job 👍

Jokinen:

PR #299, opened 2020-09-17
KK-564 | Fix redirect to registration for unregistered users

Jokinen:
At some point unregistered users have no longer been redirected to registration after logging in.

After this change the `useProfile` hook checks if a signed in user has a profile. If not, the user is redirected to being the registration process.

The menu also will only show a logout button when the user is logged in but does not yet have a profile.


tuovinensanttu (approved):

Jokinen:
Rebased in order to resolve merge conflicts.

PR #300, opened 2020-09-18
KK-563 | Fix child list not updating after adding a new child

Jokinen:
The child list used data from the redux store. The child add modal
relied on Apollo's refetchQueries to update related data. However,
refetched queries are not at any point synced into the redux store.

This behaviour may have broken in 054e284 where we changed the
synchronising mechanism between Apollo and redux from useEffect to
a callback. The callback is only fired when that particular hook makes
a query. The data the hook returns in turn changes when the cached data
changes. In this way the useEffect approach will more consistently keep
the two states in sync.

Using useEffect in such a way is an anti-pattern. Apollo likely
provides some other mechanism which allows us to subrscibe to state
changes, but I doubt that they are worth it when we can just as well
use source the data from Apollo's state.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 # 1.4.0
 
+### Fixed
+
+- Child list not showing changes without a refresh
+
 ### Changed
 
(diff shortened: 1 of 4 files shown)


tuovinensanttu (approved):
LGTM and list is updated when new child is added.

PR #302, opened 2020-09-21
KK-567 | Handle child already enrolled error

Jokinen:
When the user attempts to enrol a child into an event the child is already enrolled in they will be redirected to the event page and a warning notification will be shown to them.

### Reproduction

While logged in into a kukkuu account that has a child

1) Go to child page
2) Go to event page
3) Select occurrence
4) Copy URL
5) Paste URL into a new tab
6) In the first tab, enrol
7) in the second tab, enrol

(In the second tab) Expect to be redirected into the event page and see a warning toast.

### DIFF
--- a/src/domain/event/enrol/Enrol.tsx
+++ b/src/domain/event/enrol/Enrol.tsx
@@ -1,4 +1,4 @@
-import React, { FunctionComponent } from 'react';
+import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useHistory, useParams } from 'react-router-dom';
 import { useMutation, useQuery } from '@apollo/react-hooks';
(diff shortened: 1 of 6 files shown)


src/domain/event/enrol/Enrol.tsx
```diff
@@ -24,7 +25,20 @@ import { saveChildEvents, justEnrolled } from '../state/EventActions';
...
+function check<D>(data: D, resultFactory: (data: D) => boolean): boolean {
+  return resultFactory(data);
+}
+
+function containsAlreadyJoinedError(
+  errors: ReadonlyArray<GraphQLError>
+): boolean {
+  return errors.some(
+    (error: GraphQLError) =>
+      error.extensions?.code === 'CHILD_ALREADY_JOINED_EVENT_ERROR'
```
tuovinensanttu, line 37: Maybe store this error code in constant's file? I think similar thing was done in `profile-ui`.

tuovinensanttu (approved):
Looking good

Jokinen:
Rebased

PR #311, opened 2020-10-02
KK-513 | Fix free spot notification subscriptions

Jokinen:
## Description

This PR contains code changes which try to generalise the notification control button enough so that it can be shared between the enrolment and occurrence list pages. Currently the component contains the subscription, unsubscription and subscription confirmation modal components. Depending on the status of the user's subscription, one of the two buttons is shown. Because the confirmation modal should always be shown when subscribing, I thought that it'd be convenient to also include it into this component. The downside is that the modal is declared multiple times, although that shouldn't be a huge caveat.

I also refactored the tests I previously implemented tests with `enzyme` that I now replaced with `testing-library`, because `enzyme` is inconvenient for integration level test cases.

## Context

In this PR I make fixes to the behaviour of the KK-513 PR which
* Fix issues that were noted during review
* Address KK-560

**Issues noticed during review**
The occurrence listing page had a few design details which were wrong
* When an occurrence is full, the number in the "free places" column should be replaced with the text "FULL" (_see figure 1_)
* The button for subscribing should not contain the information about the events capacity status. E.g., instead of the label "Full - Subscribe to notifications", the button should be labeled "Subscribe to notifications" (_see figure 2_)
* When subscribing from the occurrence list view, no confirmation modal was shown to the user (_see figure 3_)

**KK-560**
This PR also handles KK-560. The ticket in question was tracked in order to ensure that the user experience when trying to enrol into a full event is optimised. Previously the user was shown a generic error message and they weren't able to recover from the error.

Now the user no longer sees a generic error message. Instead the page (enrolment page, figure 5) is refreshed, in which case it will update to reflect that the event has become full. This can occur when the event becomes full between the time the page has loaded and the user finishes the enrolment procedure.

[KK-560](https://helsinkisolutionoffice.atlassian.net/browse/KK-560)
[KK-513](https://helsinkisolutionoffice.atlassian.net/browse/KK-513)

## How Has This Been Tested?

I've extended the existing test cases to cover the user flow into the phase after which they toggle on the modal.

As a logged in user with an existing child

1. Clicks on child
2. Select "Upouusi tapahtuma" (one occurrence should be full, this may change as people play with the data)
    a. Observe that instead of 0 free places you see "FULL". Observer that
    b. Observe that instead the button label does not contain the word "Full"
3. Click on "Subscribe to notifications"
    a. Expect a confirmation dialog to show
    b. Expect for it to fill criteria for accessibility
4. On confirmation modal, click subscribe
   a. Expect for modal to close
   b. Expect notification subscription button to change into unsubscription button
5. Click on unsubscription button
   a. Expect button to change into subscribe button

KK-560 can't be conveniently tested without access to the admin and therefore I won't provide instructions. As general context, the tested has to access the enrolment page for an occurrence, do nothing there, go to the admin and set the capacity override so that the occurrence becomes full, and then enrol.

## Screenshots

Screenshot 2020-10-02 at 9 07 47

_Figure 1_

Screenshot 2020-10-02 at 9 09 57

_Figure 2_

Screenshot 2020-10-02 at 9 12 15

_Figure 3_

### DIFF
--- a/src/domain/event/EventOccurrenceNotificationControlButton.tsx
+++ b/src/domain/event/EventOccurrenceNotificationControlButton.tsx
@@ -30,7 +30,7 @@ const EventNotificationControlButton = ({
   subscribeLabel,
   unsubscribeLabel,
 }: Props) => {
-  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [isModalOpen, setIsModalOpen] = useState<boolean>(false);
   const [unsubscribe] = useUnsubscribeFromFreeSpotNotificationMutation(
     eventId,
     childId
--- a/src/domain/event/EventOccurrenceNotificationControlButton.tsx
+++ b/src/domain/event/EventOccurrenceNotificationControlButton.tsx
@@ -1,0 +1,107 @@
+import React, { useState } from 'react';
+
+import Button from '../../common/components/button/Button';
+import EventNotificationSubscriptionModal, {
+  Occurrence,
+} from './EventNotificationSubscriptionModal';
+import useUnsubscribeFromFreeSpotNotificationMutation from './useUnsubscribeFromFreeSpotNotificationMutation';
+import styles from './eventOccurrenceNotificationControlButton.module.css';
+
+type Props = {
+  childId: string;
+  eventId: string;
+  isSubscribed: boolean;
+  occurrence: Occurrence;
+  onUnsubscribed?: () => void;
+  onSubscribe?: () => void;
+  onSubscribed?: () => void;
+  subscribeLabel: string;
+  unsubscribeLabel: string;
+};
+
+const EventNotificationControlButton = ({
+  childId,
+  eventId,
+  isSubscribed,
+  occurrence,
+  onUnsubscribed,
+  onSubscribe,
+  onSubscribed,
+  subscribeLabel,
+  unsubscribeLabel,
+}: Props) => {
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [unsubscribe] = useUnsubscribeFromFreeSpotNotificationMutation(
+    eventId,
+    childId
+  );
+
+  const handleSubscribe = () => {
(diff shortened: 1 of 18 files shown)


src/domain/event/EventOccurrenceNotificationControlButton.tsx
```diff
@@ -0,0 +1,107 @@
...
+  isSubscribed,
+  occurrence,
+  onUnsubscribed,
+  onSubscribe,
+  onSubscribed,
+  subscribeLabel,
+  unsubscribeLabel,
+}: Props) => {
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [unsubscribe] = useUnsubscribeFromFreeSpotNotificationMutation(
```
tuovinensanttu, line 34: I don't remember how strict we have been about adding types to `useState` hooks.

tuovinensanttu (approved):
Code is looking nice, one little nitty picky comment tho. 

I alse tested the UI, correct text is used if event (or is it occurrence in this case, I'm little confused by these 😅 ) is full. When subscribing, modal is shown correctly 👍 

I'm just wondering, should the modal be shown if user unsubscribes as well? Miss clicks can cause people to drop back into the end of "line".

src/domain/event/EventOccurrenceNotificationControlButton.tsx
Jokinen: Me neither, but I think that it makes sense to add one here, thanks for pointing this out 👍

Jokinen:
Subscribing to notifications doesn't put you in a real queue. Instead, it just tags you as somebody who wants to receive notifications. The order the modal references in its description, is the order in which the users presses the enrol button, after they've received a notification and used the link in it to access the enrolment page.

In that way it seems that the UX may be unclear here in a different way--it's hard for the user to understand what they are signing up for. I've communicated with Anniina about this scenario and changes won't be made for now to adjust this.

PR #274, opened 2020-07-13
KK-504 | Add a option to select languages spoken at home

badrange:
This is just a draft - first of all it depends on a PR for HDS that allows dropdown components to be multiselect with search. 

The language lists are the most spoken in Finland, we might want to drop the ones with only a couple of speakers. The most common languages are listed first, the rest should probably be listed automatically (TODO).

PR #312, opened 2020-10-12
KK-504 | Add support for languages spoken at home

Jokinen:
## Description

Adds support for the `languagesSpokenAtHome` field. Guardians can now control this information in their profiles through a multiselect combobox component. As a part of this feature, I helped the HDS team implement the `Combobox` component into HDS.

## Context

After these change users can provide information about the languages that are spoken in their homes. The Culture Kids maintainers are then able to use this data to offer events that take user languages into account better.

[KK-504](https://helsinkisolutionoffice.atlassian.net/browse/KK-504)
Closes #274 

## How Has This Been Tested?

I added tests, but they don't actually cover the languages spoken at home field. In the implementation I chose to entangle data with the component for choosing languages. This means that other components using it don't have to load the data, but it also means that testing the component itself requires mocking the API. I'm not entirely sure of how strong of a sign it's to manage the data differently, but for now I did not make changes.

This functionality will be easier to cover by using `e2e` tests. It would be included in
* https://helsinkisolutionoffice.atlassian.net/browse/KK-591
* https://helsinkisolutionoffice.atlassian.net/browse/KK-600

Without logging in with a Tunnistamo account that's not registered to Kukkuu

1. Open home page
2. Click register
3. Fill short form
4. Authenticate in Tunnistamo
5. Fill long form
    - Expect to be able to control languages spoken at home
    - Feel free to test the combobox in the way you feel appropriate. There are many usecases for it of which many can't likely be fixed in this PR
6. Submit registration form
7. Click Edit information
8. Expect to see languages spoken at home filled with the values you chose previously
9. Expect to be able to edit those values


tuovinensanttu (approved):
Lot of going on here, but imo everything looks good!

PR #313, opened 2020-10-13
KK-585 | Handle unhandled error codes

Jokinen:
## Description

Previously two error messages were not handled for subscriptions: `ALREADY_SUBSCRIBED_ERROR` and `OCCURRENCE_IS_NOT_FULL_ERROR`. This PR add error handling logic which should help the user recover from these errors.

## Context

After this change users facing the `ALREADY_SUBSCRIBED_ERROR` no longer see a generic error message. Instead the page is refreshed which should update stale data and show, that what they want, to subscribe to notifications, is already taking place. In other words we don't directly inform the user about the error at all.

`OCCURRENCE_IS_NOT_FULL_ERROR` can take place when the user attempts to subscribe to notifications for an event that has free places left. This can potentially happen when user 1 loads the page and leaves it open for a while, and in the meanwhile user 2 cancels their enrolment. In these instances we show an error message that informs the user about a changed capacity.

[KK-585](https://helsinkisolutionoffice.atlassian.net/browse/KK-585)

## How Has This Been Tested?

I've tested this by hand. Testing this with automated tests would require mocking the GraphQL API which isn not ideal.

_As a user subscribing to notifications of an occurrence I have already subscribed to, I do not want to see an error message, because everything is as I want it to be._

1. As a registered logged in user in your god children profile
2. Select a child
3. Select an event
4. Ensure that there's a full occurrence you can subscribe to 
     - if there are no full occurrences, change capacity from admin, or subscribe with another account first
6. Open the same view in another tab
7. Click on the subscribe to notification button
8. Go to second tab and click on the subscribe to notifications button
9. Expect no error to be shown and the button to change, reflecting the fact that you are subscribed

_As a user subscribing to notifications of an occurrence that's no longer full, I want to be informed that I can possibly sign up for the occurrence._

1. As a registered logged in user in your god children profile
3. Select a child
5. Select an event
7. Ensure that there's a full occurrence 
    - if there are no full occurrences, change capacity from admin, or subscribe with another account first
11. Change capacity from admin, or unsubscribe with another account
12. Click the subscribe to notifications button
13. Expect to see an error that guides the user to refresh the page

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
 - Show no other button than log out for logged in unregistered users
 - Update accessibility statement
 - Show events as upcoming events also for a while after they have started
+- No longer display an error when a user subscribes to notifications for an occurrence they have already subscribed to, instead refresh the page to fetch current most data
+- Display a more specific error message when a user subscribes to notification for an occurrence which has available capacity
 
 ### Fixed
 
--- a/src/common/translation/i18n/en.json
+++ b/src/common/translation/i18n/en.json
@@ -158,6 +158,7 @@
     },
     "enroll": {
       "error": {
+        "capacityHasChanged": "The available capacity of the event has changed. Please refresh your browser.",
         "childAlreadyJoined": "This child has already been enrolled in this event."
       },
       "buttonText": "Register for event"
--- a/src/common/translation/i18n/fi.json
+++ b/src/common/translation/i18n/fi.json
@@ -158,6 +158,7 @@
     },
     "enroll": {
       "error": {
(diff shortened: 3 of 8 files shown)


tuovinensanttu (commented):
Both manual testing cases work like you described. Although on the first case, cancelling the subscription in the same situation still shows an error. Is that intentional?

Jokinen:
The error codes we are adding support for are specific to subscribing. I don't think the generic error is intentional per se, but it could be thought of as something we haven't gotten around to doing yet. I added a task for this: https://helsinkisolutionoffice.atlassian.net/browse/KK-598

tuovinensanttu (approved):
LGTM

PR #320, opened 2020-11-17
KK-642 | Fix error caused by missing ResizeObserver

Jokinen:
## Description

Fixes the error caused by missing ResizeObserver in some browsers. In essence these changes take a newer version of HDS into use. This version contains a ponyfill for ResizeObserver which should resolve our issue.

## Context

[KK-642](https://helsinkisolutionoffice.atlassian.net/browse/KK-642)

## How Has This Been Tested?

I've manually tested the key views with HDS components in order to quickly check that there are no regressions. I haven't tested this with a browser that doesn't support `ResizeObserver` because I couldn't find an easy approach to do so.

Based on my research it's too difficult to set up an environment to test this. Unless you know better, or happen to have an iPhone with an older version of iOS, I can't provide viable testing instructions.


tuovinensanttu (approved):
LGTM 👍

PR #324, opened 2020-11-19
KK-659 | Fix profile link being visible without authentication

Jokinen:
## Description

Hides the link to user's profile on the home page when the user is not authenticated.

The application stores the user's profile into redux and persists it into local storage. When the user logs out, this data is cleared. When the user's authentication expires, this data is not cleared. For now I did not add the logic for clearing this data. The current architecture creates a cyclical dependency between the `userManager` and the redux store.

## Context

[KK-659](https://helsinkisolutionoffice.atlassian.net/browse/KK-659)

## How Has This Been Tested?

I've added a unit test that checks that the button is not visible when the user is not authenticated.

Login to the application

1. Wait until your authentication expires
1. Instead of following the instructions of the modal, change the address to `/` by hand
1. Expect to not see a button pointing to the profile page


tuovinensanttu (approved):
LGTM

Jokinen:
Rebased on current develop.

PR #325, opened 2020-11-19
KK-661 | Do not browser validate child form

Jokinen:
## Description

Adds the `noValidate` attribute to the child form. This ignores browser validation so that it doesn't block our custom validation.

I also refactored the view to use `<Form />` from `formik` which automatically binds the `onSubmit` and `onReset` events.

## Context

[KK-661](https://helsinkisolutionoffice.atlassian.net/browse/KK-661)

## How Has This Been Tested?

I've tested this manually. I don't think that a meaningful test can be written to check the actual user interaction. Instead I tested that the child's form element has `noValidate` set to `true`.

As a logged in user

1. Click add a new child
1. Fill a the birthday 02-02-2010
1. Fill other required fields
1. Submit
1. Expect to see a view where you are being told that your child is not eligible

### DIFF
--- a/src/domain/child/form/ChildForm.tsx
+++ b/src/domain/child/form/ChildForm.tsx
@@ -1,5 +1,5 @@
 import React, { FunctionComponent } from 'react';
-import { Formik, FieldArray, FormikErrors, FormikProps } from 'formik';
+import { Formik, FieldArray, FormikErrors, FormikProps, Form } from 'formik';
 import { useTranslation } from 'react-i18next';
 import classnames from 'classnames';
 import * as yup from 'yup';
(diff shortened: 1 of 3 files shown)


tuovinensanttu (approved):

Jokinen:
Rebase onto current develop.

src/domain/child/form/ChildForm.tsx
```diff
@@ -121,12 +121,12 @@ const ChildForm: FunctionComponent<ChildFormProps> = ({
...
       {({
         isSubmitting,
-        handleSubmit,
         values,
         errors,
         touched,
       }: FormikProps<ChildFormValues>) => (
-        <form onSubmit={handleSubmit} id="childForm">
+        <Form id="childForm" noValidate>
+          he
```
badrange, line 129: Is this "he" a stray line of code?

Jokinen: Yes it is 💆‍♂️ I'll add a fix tomorrow.

Thanks for pointing it out👍

badrange: Glad I could be of help! 

I'm following the changes in the repo because I know I have a lot to learn from how you are developing it further. Congrats with browser testing coming in!!

tuovinensanttu: Nice catch! Seems like I need a pair of glasses 😒

PR #326, opened 2020-11-19
KK-660 | Add temporary link for giving feedback

Jokinen:
## Description

Adds a link to a feedback form that will be removed at some point. I needed to pull button styles from `hds-core` because I think that there's no link button in HDS.

## Context

[KK-660](https://helsinkisolutionoffice.atlassian.net/browse/KK-660)

## How Has This Been Tested?

I've tested this by hand.

Expect to see a link-button in profile and child profile pages

## Screenshots

Screenshot 2020-11-19 at 14 03 00
Screenshot 2020-11-19 at 14 03 09

### DIFF
--- a/src/common/components/giveFeedbackButton/GiveFeedbackButton.tsx
+++ b/src/common/components/giveFeedbackButton/GiveFeedbackButton.tsx
@@ -1,0 +1,31 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import 'hds-core/lib/components/button/button.css';
+
+import styles from './giveFeedbackButton.module.scss';
+
+const feedbackFormLink =
+  // eslint-disable-next-line max-len
+  'https://docs.google.com/forms/d/e/1FAIpQLSdqw2Lq3qooEeRdgr0sV0-Wv-4XcV7IZVzq1HuWoLRa2M7tEg/viewform?usp=pp_url&entry.1982410290=Kulttuurin+kummilapset';
+
+const buttonClassName = [
+  'hds-button',
+  'hds-button--secondary',
+  'hds-button__label',
+  'hds-button--theme-black',
+  styles.buttonLink,
+].join(' ');
+
+const GiveFeedbackButton = () => {
+  const { t } = useTranslation();
+
+  return (
+    <div className={styles.container}>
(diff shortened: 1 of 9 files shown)


src/common/components/giveFeedbackButton/GiveFeedbackButton.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import 'hds-core/lib/components/button/button.css';
+
+import styles from './giveFeedbackButton.module.scss';
+
+const feedbackFormLink =
+  // eslint-disable-next-line max-len
+  'https://docs.google.com/forms/d/e/1FAIpQLSdqw2Lq3qooEeRdgr0sV0-Wv-4XcV7IZVzq1HuWoLRa2M7tEg/viewform?usp=pp_url&entry.1982410290=Kulttuurin+kummilapset';
```
tuovinensanttu, line 9: Even though this feedback implementation is temporary, would it be better if it added in the excel instead? If something happens and it needs to be replaced it would be little easier.

tuovinensanttu (approved):
Looking good!

src/common/components/giveFeedbackButton/GiveFeedbackButton.tsx
Jokinen: My concern with adding this link to the sheet is that once this component is removed, the sheet is not updated and will continue to contain this link. Having the link within this component will make the maintenance chore of removing it easier.

I'm going to confirm that we'll only provide this link in Finnish. If not, I'll manage it through the sheet.

Jokinen: When it comes to replacing it, I think that in this scenario the effort is roughly the same. Either the developer finds the correct file and copy pastes the link directly into the source, or they copy paste it into the sheet and pull it with the script. Generally I'd be in favour of storing the "copy" in the same place, so in the sheet, even if it took more effort.

Jokinen: We are going to support English speakers as well, so the link will be moved to the sheet :)

PR #335, opened 2020-12-14
Use Github Actions for CI/CD instead of Gitlab 

joonvena:
## Context

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -1,0 +1,30 @@
+name: Build & Production
+on:
+  push:
+    branches:
+      - master
+
+env:
+  CONTAINER_REGISTRY: ghcr.io
+  CONTAINER_REGISTRY_USER: ${{ secrets.GHCR_CONTAINER_REGISTRY_USER }}
+  CONTAINER_REGISTRY_PASSWORD: ${{ secrets.GHCR_TOKEN }}
+  CONTAINER_REGISTRY_REPO: ghcr.io/city-of-helsinki/${{ github.event.repository.name }}
+  REPO_NAME: ${{ github.event.repository.name }}
+  KUBECONFIG_RAW: ${{ secrets.KUBECONFIG_RAW_STABLE }}
+  BUILD_ARTIFACT_FOLDER: 'build_artifacts'
+  SERVICE_ARTIFACT_FOLDER: 'service_artifacts'
+  SERVICE_PORT: 80
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Build
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build
+        uses: andersinno/kolga-build-action@v2
+        env:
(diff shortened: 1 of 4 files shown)


.github/workflows/stop_review.yml
```diff
@@ -0,0 +1,19 @@
...
+  stop_review:
+    runs-on: ubuntu-latest
+    name: Stop Review
+    steps:
+      - uses: andersinno/kolga-setup-action@v2
+
+      - name: Stop Review
+        uses: andersinno/kolga-review-cleanup-action@v2
+        with:
+          namespace: ${{ env.K8S_NAMESPACE }}
```
Jokinen, line 19: Missing newline.

.github/workflows/production.yml
```diff
@@ -0,0 +1,30 @@
+name: Build & Production
```
Jokinen, line 1: We already have a job includes the word build, but it just tests that the static build is working in general. Probably one or both should be changed to be more specific in order to avoid confusions. E.g. Build static / Build image.

Jokinen:
I accessed this URL: https://kukkuu-ui-539-dev-github-actions.test.kuva.hel.ninja/

I don't think that this is the correct url, because it didn't seem to work. I can't remember what Rurik typed in the demo.

I faced the same error as with the previous version of the infra, the certification for the environment does not work ("issuer unknown"). Firefox allows me to bypass this warning, but I ended up seeing a `404 error`, which leads me to believe that I did not have the correct url.

joonvena:
We are using subdomain for the new cluster in development stage. You can access the deployment:

https://kukkuu-ui-539-dev-github-actions.newtest.kuva.hel.ninja/fi/home

Jokinen:
Nice, it works! The certificate issue is still there though and Chrome doesn't allow the page to be opened. Can you confirm that this is expected.

Screenshot 2020-12-15 at 8 36 41

Also, can you give talk a bit about the timeframe for these changes. In other words, should these be merged as soon as possible, or should we still wait for something?

joonvena:
This cert error is expected in the review environment. There is more recreation of certificates / environments happening when doing the review deployment and lets-encrypt has a limit for the certificate requests (I think it will issue 4 times in a week). So this is more like a safety measure for the lets-encrypt. Certificates are valid in staging and prod. We will most likely merge these changes today. It can cause little outage in the QA as it will trigger the Gitlab pipeline also when we make the push. We will remove the mirroring after we have the production cluster setup and updated also.

Jokinen:
Ok, thank you for the information! This was helpful.

.github/workflows/production.yml
joonvena: Is it ok if you change the existing static build test workflow name? As we have implemented these workflows for all projects using this naming so it would be great if all projects had unified naming.

.github/workflows/stop_review.yml
joonvena: Fixed

PR #338, opened 2020-12-17
KK-686 | Fix event group events query

Jokinen:
## Description

Adjust event group events query to only show events that are available to the child.


tuomas777 (approved):
Yes 👍

PR #341, opened 2021-01-19
Chore | Use GitHub flow instead of GitFlow

Jokinen:
## Description

Transitions this repository into using GitHub flow that is recommended in best practices.

Because a new version was just release, and the development and master branches are in sync as content goes, this is an easy time to make the transition.

## Context

[Best Practices: Version Control](https://dev.hel.fi/version-control#git-workflow)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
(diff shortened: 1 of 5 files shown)


tuovinensanttu (approved):
LGTM 👍

PR #347, opened 2021-02-05
Chore | Run browser tests against PR env

Jokinen:
## Description

Runs browser tests against the PR environment.

## Context

After this change developers will get more instant feedback from possible breakages in browser tests when it comes to their changes.


tuovinensanttu (approved):
👍

PR #354, opened 2021-02-18
KK-682 | Fix children list screen reader content

Jokinen:
## Description

Use same content for screen reader as for visual users.

Removes aria-label from children in child list so that the content of the button is read aloud with screen readers.

Refactors browser test because the change caused some selector patterns to become unusbale.

## Context

[KK-682](https://helsinkisolutionoffice.atlassian.net/browse/KK-682)

## How Has This Been Tested?

With VoiceOver and Safari.

Expect screen reader users to have access to all the information visual users do.

### DIFF
--- a/browser-tests/childrenFeature.ts
+++ b/browser-tests/childrenFeature.ts
@@ -33,7 +33,7 @@ function buildEditChild() {
   };
 }
 
-const childName = /Hertta Citron .*/;
+const childName = /.* Hertta Citron .*/;
 
 fixture`Children feature`
   .page(route())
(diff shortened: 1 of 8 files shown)


browser-tests/childrenFeature.ts
```diff
@@ -117,8 +120,8 @@ test('As a guardian I want to add and delete a child', async (t) => {
   await t.wait(1000); // 1s
 
   // Assert that the child can be found
-  await t.expect(godchildrenProfilePage.child(newChildName).exists).ok();
+  await t.expect(godchildrenProfilePage.child(newChildNameRegepx).exists).ok();
```
tuovinensanttu, line 123: This line causes browser test to fail.

Jokinen: Hmn, bit of an unfortunate issue because it seems that it has to do with the wait time. The screenshot of the original error contained the button for the created child. I re-ran the suite now and it passed. My guess is that not a lot of extra time is needed, so I added 0.5s.

tuovinensanttu (approved):

PR #355, opened 2021-02-18
Possibility to run workflows manually

Hi-Fi:
## Description

Allows running just deployment step manually within workflows (e.g. update parameters or make redeploy for some reason)

## Context

[DEVOPS-256](https://helsinkisolutionoffice.atlassian.net/browse/DEVOPS-256)

### DIFF
--- a/.github/workflows/stop_review.yml
+++ b/.github/workflows/stop_review.yml
@@ -2,6 +2,13 @@ name: Stop Review
 on:
   pull_request:
     types: [closed]
+  workflow_dispatch:
+    inputs:
+      head_ref:
+        description: "Branch where review environment was created from (as in GitHub). Has to give as branch is probably removed and not available at the list."
+        required: true
+      pr_number:
+        description: "Pull request number of the environment creator"
 
 env:
   KUBECONFIG_RAW: ${{ secrets.KUBECONFIG_RAW }}
@@ -12,6 +19,9 @@ jobs:
     name: Stop Review
     steps:
       - uses: andersinno/kolga-setup-action@v2
+        with:
+          head_ref: ${{ github.event.inputs.head_ref }}
+          pr_number: ${{ github.event.inputs.pr_number }}
 
       - name: Stop Review
         uses: andersinno/kolga-review-cleanup-action@v2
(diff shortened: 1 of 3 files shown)


Hi-Fi:
Can't be tested before workflows end up to default branch.

.github/workflows/stop_review.yml
```diff
@@ -2,6 +2,13 @@ name: Stop Review
 on:
   pull_request:
     types: [closed]
+  workflow_dispatch:
+    inputs:
+      head_ref:
+        description: "Branch where review environment was created from (as in GitHub). Has to give as branch is probably removed and not available at the list."
```
Jokinen, line 8:
```suggestion
        description: "Branch where review environment was created from (as in GitHub). Has to be given as the branch is probably removed and no longer available in the branch list."
```

Jokinen (approved):
Looks good. 

I'm wondering how this works. As far as I understand this would allow us to more easily configure settings (after we change the repo into using runtime configurations).

Do I understand correctly that the intended use for updating for instance the production configs would be to open a new branch with the fixed configurations, and then manually running the production deployment job with that branch?

As I understand this change also allows PR envs to be stopped on demand in the case you do not need it, or in case one has failed to stop?

Hi-Fi:
> Looks good.
> 
> I'm wondering how this works. As far as I understand this would allow us to more easily configure settings (after we change the repo into using runtime configurations).
> 
> Do I understand correctly that the intended use for updating for instance the production configs would be to open a new branch with the fixed configurations, and then manually running the production deployment job with that branch?
> 
> As I understand this change also allows PR envs to be stopped on demand in the case you do not need it, or in case one has failed to stop?
> 
> The browser test is still really finnicky with the first authentication routine even with more added wait time. It's unrelated with this PR.

I'll create demo app to test this PR, as it might have some glitches which can't be checked before things are in default branch.

Main idea is to allow running of stop-review if environment is not needed or it has failed remove for some reason. The build related triggers are bonus to be able to redeploy app without need to build every time (if changes are coming from secrets).

terovirtanen:
Workflows have been removed, DevOps pipelines are used instead

PR #356, opened 2021-02-18
KK-684 | Fix incorrect tab order in occurrence list

Jokinen:
## Description

Fixes incorrect tab order in occurrence list when using a mobile sized screen.

Implements a custom view for mobile so that table headers and rows match.

## Context

[KK-684](https://helsinkisolutionoffice.atlassian.net/browse/KK-684)

## How Has This Been Tested?

I've used VoiceOver and Safari to test this.

One thing worth noting is that the technical implementation uses table, but override's its display values in which case the screen reader is not able to see a table.

The layout on mobile looks like it could be difficult to reproduce using tables do I haven't attempted to fix the semantics here.

### DIFF
--- a/src/domain/event/EventOccurrence.tsx
+++ b/src/domain/event/EventOccurrence.tsx
@@ -30,6 +30,12 @@ const EventOccurrence = ({ occurrence }: EventOccurrenceProps) => {
     `ddd ${DEFAULT_DATE_FORMAT}`
   );
   const time = formatTime(newMoment(occurrence.time), DEFAULT_TIME_FORMAT);
+  const machineReadableDate = formatTime(newMoment(occurrence.time));
+  const machineReadableTime = formatTime(
+    newMoment(occurrence.time),
+    'HH:mm:ss.SSS'
+  );
+  const machineReadableDateTime = `${machineReadableDate} ${machineReadableTime}`;
 
   const hasCapacity = Boolean(
     occurrence.remainingCapacity && occurrence.remainingCapacity > 0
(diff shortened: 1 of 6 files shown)


Jokinen:
@tuovinensanttu this has probably dropped off of your todo list.

src/domain/event/EventOccurrence.tsx
```diff
@@ -30,53 +30,73 @@ const EventOccurrence = ({ occurrence }: EventOccurrenceProps) => {
...
+        <td className={styles.remainingCapacity}>{remainingCapacity}</td>
+        <td className={styles.occurrenceSubmit}>{submitCell}</td>
+      </tr>
+      <tr className={styles.occurrence}>
+        <td className={styles.occurrenceDate}>{date}</td>
+        <td className={styles.occurrenceTime}>{time}</td>
+        <td className={styles.occurrenceVenue}>{occurrence.venue.name}</td>
+        <td className={styles.remainingCapacity}>{remainingCapacity}</td>
+        <td className={styles.occurrenceSubmit}>{submitCell}</td>
+      </tr>
```
tuovinensanttu, line 98: These styles are commented out.

src/domain/event/eventOccurrence.module.scss
```diff
@@ -32,31 +56,36 @@ tr {
...
-    grid-column: 1 / span 4;
-  }
+
+  // td.remainingCapacity {
+  //   grid-column: 4;
+  //   grid-row: 1 / span 2;
+  //   justify-self: end;
+  //   align-self: center;
+  // }
+  // td.occurrenceVenue {
```
tuovinensanttu, line 72: Is there a use for these in the future?

tuovinensanttu (approved):

PR #357, opened 2021-02-22
KK-683 | Read titles aloud

Jokinen:
## Description

This change consists of two components--"manually" announcing title changes and changing the title as soon as possible.

The manual announcement trigger has been done by hooking into the react-helmet's event logic. When the title is changed, it's announced with an aria-live region.

The authentication process has been adjusted. Instead of wrapping the application into a single loading spinner, each page has been wrapped in one. By doing this we allow some of the logic within the page to update, in this case the logic that changes the title. This in turn allows the title to be changed before the content of the page has loaded, decreasing the chance of a "double title" announcement on a first page load.

## Context

[KK-683](https://helsinkisolutionoffice.atlassian.net/browse/KK-683)

## How Has This Been Tested?

I've used Safari and VoiceOver to test this. For me, VoiceOver announces the title change once, and then the manually trigger does it a second time. When I consulted accessibility professionals, they said that screen readers do not read aloud titles--I assume that I have some configuration wrong.

Test that title changes are announced on page changes.


tuovinensanttu (approved):
Really nice implementation! Tested this with:
1) VoiceOver + Chrome => working nicely, didn't notice that titles would be read twice
2) NVDA + Chrome => titles were read aloud as well!

Jokinen:
Thank you for checking with NVDA 👍

PR #365, opened 2021-03-26
KK-739 | Add corona virus info to the home page

Jokinen:
## Description

Adds a coronavirus info text on the home page of the service.

I made this info configurable so that it can be removed without extensive code side changes.

I also fixed the typography on the home page so that the hierarchy intended in the design actually works.

## Context

[KK-739](https://helsinkisolutionoffice.atlassian.net/browse/KK-739)
[Design](https://app.abstract.com/projects/447ded2e-2949-4a2b-a67e-874ab853cf1d/branches/master/commits/latest/files/5CBA3ABA-4F29-45FD-888F-B3121341F6AA/layers/C5995A53-78AA-4A8C-9497-A46BB062B9C3?collectionId=3a78ab05-980d-448d-841d-8ea24b6bf5bc&collectionLayerId=75d34585-2f77-42cd-b581-f767fc230bd9&mode=build&selected=root-ED7E326C-82D1-4DF3-A682-EACA24ABB1C1)

## Screenshots

Screenshot 2021-03-26 at 9 17 39
Screenshot 2021-03-26 at 9 17 45


src/domain/__tests__/config.test.js
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  afterEach(() => {
+    process.env = env;
+  });
+
+  describe('featureFlagShowCoronaInfo', () => {
+    it('should return true when REACT_APP_FEATURE_FLAG_SHOW_CORONAVIRUS_INFO is true', () => {
+      process.env.REACT_APP_FEATURE_FLAG_SHOW_CORONAVIRUS_INFO = 'true';
+      process.env.REACT_APP_FEATURE_FLAG_SHOW_CORONAVIRUS_INFO = true;
```
tuovinensanttu, line 17: The first one could be removed.  Or is this following the same logic as the test below and is just missing 
`expect(Config.featureFlagShowCoronaInfo).toEqual(true);` in between these two lines?

tuovinensanttu (approved):
Looks good to me!

src/domain/__tests__/config.test.js
Jokinen: I'm trying to check that both a boolean and a string value yield true, so yup, I'm missing one expect 👍

src/domain/home/hero/HomeHero.tsx
```diff
@@ -25,7 +27,13 @@ const HomeHero = ({
       <div className={styles.heroContainer}>
         <div className={styles.hero}>
           <h1>{t('appName')}</h1>
-          <p> {t('homePage.hero.descriptionText')}</p>
+          <p className={styles.bodyXl}>{t('homePage.hero.descriptionText')}</p>
+          {Config.featureFlagShowCoronaInfo && (
```
badrange, line 31: What if you used a generic name for this, for example `showFrontPageAlert` or something like that? Then it can be reused for other things in the future. 

I'm watching this project because I can learn from the improvements you're making - one of the advantages of open source!

Jokinen: To make the notification at least somewhat reusable would be relatively difficult when contrasted with the scope of this task. Naming the feature flag in a generic way could be confusing. First, it would be difficult to say what notification it would toggle. Second, it would imply functionality that's not implemented. Using a name that's not reflective of the current state, but instead a future vision, can be really confusing.

Moreover, the feature flag is a stretch in scope in itself and only meant as an escape hatch in case there's not enough development bandwith to remove the notification when events are resumed.

Thank you for the suggestion!

badrange: Everything you say makes sense! I do hope you'll be able to toggle it of and remove it soon..

PR #366, opened 2021-03-29
Capture network errors into Sentry when authenticating

Jokinen:
## Description

Users are receiving unexpected network errors when registering. We suspect that the error fails to be passed into Sentry at this point in code.


tuomas777 (approved):
👍

PR #370, opened 2021-04-08
Redirect from www to apex domain

Hi-Fi:
Jira: DEVOPS-306

Simplifies integration configuration

## Context

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -45,6 +45,17 @@ jobs:
       - uses: actions/checkout@v2
       - uses: andersinno/kolga-setup-action@v2
 
+      - name: Set multiline ingress annotation to environment variables
+        env:
+          WWW_REDIRECT: |
+            if (\$host = 'www.kummilapset.hel.fi' ) {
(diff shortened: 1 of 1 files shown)


Jokinen (commented):
Should we also set additional hostnames like here: https://github.com/City-of-Helsinki/kuva-demo-project/pull/11/files#diff-7bb7a67ed05f3845ed1f1b92dc4b87dfb0dd9aba208cca5489055ccef315117bR104

Hi-Fi:
> Should we also set additional hostnames like here: https://github.com/City-of-Helsinki/kuva-demo-project/pull/11/files#diff-7bb7a67ed05f3845ed1f1b92dc4b87dfb0dd9aba208cca5489055ccef315117bR104
> 
> In other words, currently the additional hostnames variable does not reference the www prefixed address. Is that correct?

Good catch, the URL referred was that .ninja URL, so the redirect has to input there manually. Those apex and www urls are just part of service URLs list given, so easier to just type manually.

Jokinen (approved):

PR #376, opened 2021-05-27
Disable front page corona info text

tuomas777:
Executed in a bit of a hurry. Needs to be done properly when there is more time.


tuovinensanttu (approved):
LGTM

PR #379, opened 2021-06-28
Chore | Update unsecure dependencies

Jokinen:
## Description

Update reac-scripts and some other dependencies in order to resolve security warnings.

Some advisories were quite old and did not have a branch 3.x.x fix for `react-scripts`. I had to re-create the lockfile in order to force the resolution of some security issues.


tuovinensanttu (approved):
👍

PR #381, opened 2021-06-30
KK-761 | Add support for Ticketmaster ticketing system

Jokinen:
## Description

Adds support for Ticketmaster ticketing system.

Events which have been configured with the Ticketmaster ticketing system will use the redirect flow for reservations.

This means that users will not sign up for the event within kukkuu. Instead, they will be notified that they will be redirected to the Ticketmaster service.

## Context

[KK-761](https://helsinkisolutionoffice.atlassian.net/browse/KK-761)

### DIFF
--- a/src/common/components/text/Text.tsx
+++ b/src/common/components/text/Text.tsx
@@ -1,0 +1,53 @@
+import React, { ReactNode, HTMLAttributes } from 'react';
+
+import styles from './text.module.scss';
+
+type TextVariant =
+  | 'h1'
+  | 'h2'
+  | 'h3'
+  | 'h4'
+  | 'h5'
+  | 'body'
+  | 'body-l'
+  | 'body-xl';
+
+type Props = {
+  as?: string | React.ComponentType<HTMLAttributes<HTMLElement>>;
+  variant?: TextVariant;
+  children: ReactNode;
+  className?: string;
+  role?: string;
+};
+
+function getElement(variant: TextVariant) {
+  switch (variant) {
+    case 'h1':
+      return 'h1';
+    case 'h2':
+      return 'h2';
+    case 'h3':
+      return 'h3';
+    case 'h4':
+      return 'h4';
+    case 'h5':
+      return 'h5';
+    case 'body':
+    case 'body-l':
+    default:
+      return 'p';
+  }
+}
+
+function Text({ as, variant = 'body', children, className, ...rest }: Props) {
+  return React.createElement(
+    as || getElement(variant),
+    {
+      className: [styles.text, styles[variant], className].join(' '),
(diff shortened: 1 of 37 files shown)


src/domain/event/EventOccurrenceRedirect.tsx
```diff
@@ -0,0 +1,140 @@
...
+            variant="secondary"
+            to={getEventPathname(location.pathname)}
+            className={styles.linkButton}
+          >
+            {t('eventOccurrenceRedirectPage.back')}
+          </LinkButton>
+          <AnchorButton href={ticketmasterUrl}>
+            {t('eventOccurrenceRedirectPage.continue')}
+          </AnchorButton>
+        </div>
```
tuovinensanttu, line 134: On smaller screen these buttons being on the same line feels little cramped. I haven't seen the designs for this, but in my opinion it would be better if they were under each other and had full width.

tuovinensanttu (approved):
LGTM 👍

tuomas777 (approved):
Very great!! 💯 

A couple of details:

- we probably don't want to show the number of free places of Ticketmaster events. The design also has it that way, but that change is unfortunately missing from the ticket
- the past event text "Please take into consideration that tickets which have been delivered through Ticketmaster..." should not be always visible. There are currently no clear specs about that, most probably it should be shown only for children of certain projects (years). Of course this can be merged now and conditional hiding added later, we just need to make sure we remember to do that 🙂

src/domain/event/EventOccurrenceRedirect.tsx
Jokinen: There was no mobile version of the design. I'll make this change.

src/common/components/text/Text.tsx
```diff
@@ -0,0 +1,53 @@
...
+    case 'h3':
+      return 'h3';
+    case 'h4':
+      return 'h4';
+    case 'h5':
+      return 'h5';
+    case 'body':
+    case 'body-l':
+    default:
+      return 'p';
```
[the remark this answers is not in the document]
Jokinen: Codewise they don't do much, but for humans they tell that those two variants are accounted for.

Jokinen:
I'll remove the ticket count column.

Personally I'm not sure if we can't just hardcode the text 🤔 It doesn't seem too harmful. But yeah, it's something we can't show conditionally even if we wanted to so it can't be resolved now.

Jokinen:
Thank you for the comments! I adjusted the mobile layout and dropped the unnecessary ticket count

Screenshot 2021-07-05 at 10 55 32
Screenshot 2021-07-05 at 10 55 16

PR #384, opened 2021-09-13
KK-773 | Prevent users from accessing event group enrolment userpath after having been enrolled into an event

Jokinen:
## Description

Introduces a check which redirects users into the profile page when there are no events to enrol into on the event group page. We assume that when there are no events, something is either wrong or the user has already enrolled.

Adds a query refetching mechanism that refetches the event group page query so that the amount of available event is correctly updated when a user enrols into an event.

### DIFF
--- a/src/domain/api/generatedTypes/eventOccurrenceQuery.ts
+++ b/src/domain/api/generatedTypes/eventOccurrenceQuery.ts
@@ -63,5 +63,5 @@ export interface eventOccurrenceQuery {
 
 export interface eventOccurrenceQueryVariables {
   id: string;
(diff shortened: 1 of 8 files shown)


tuomas777 (commented):
Seems to work great but there is still one problem: after you have unenrolled from an occurrence, you cannot access the event group view unless you reload the page. My guess would be that the unenroll mutation needs a refetch query too?

Jokinen:
Added a refetch query for the unenroll mutation 👍

tuomas777 (approved):
Splendid! 🚀

PR #391, opened 2021-10-04
Fixes to left arrow go back features.

nikomakela:
Replaced all the history.goback() -calls with hook calls that takes the user to actual desired path.

KK-753


Sartonon:
Seems like in event invites view (/fi/profile/child/:id) back button is still not working correctly

src/domain/event/route/EventRoute.tsx
```diff
@@ -1,10 +1,46 @@
...
+    skip: !!eventGroupId || !eventId,
+  });
+
+  // Go to event group page
+  if (eventGroupId) {
+    return () =>
+      history.push(
+        `/${currentLocale}/profile/child/${childId}/event-group/${eventGroupId}`
+      );
+  }
```
Sartonon, line 30: If eventGroupId exists, it basically means that we are on event group page? Maybe it could be communicated better by assigning !!eventGroupId to a variable. It is not very clear immidiately that `!!eventGroupId` means that we are on................................................. Actually I realized that I cant find a way to get eventGroupId defined, on what page exactly is this used?

Sartonon: To me this solution looks ok but I think that it just needs little bit code polishing for readability. For example when reading `!!eventGroupId || !eventId` I have hard time figuring out the actual situation when query is skipped.

nikomakela: removed the  `if (eventGroupId) -block`. Also, now returning a string instead of void.

Sartonon (approved):

PR #393, opened 2021-10-05
Improve instructions for settings up local Tunnistamo

charn:
## Description

When configuring kukkuu-ui and kukkuu-admin, this should set "Allowed applications" in Tunnistamo as expected.


nikomakela (approved):
tested and it's working! Wonderful!

PR #402, opened 2022-02-03
Pipeline yaml files and port updates

Vitals9367:
## Context


charn:
Since the port of the service is changed (and since we're still running this in a different environment) this would also need a change in configurations for the current in-use environments:
- [review](https://github.com/City-of-Helsinki/kukkuu-ui/blob/master/.github/workflows/review.yml#L14), [staging](https://github.com/City-of-Helsinki/kukkuu-ui/blob/master/.github/workflows/staging.yml#L17) and [production](https://github.com/City-of-Helsinki/kukkuu-ui/blob/master/.github/workflows/production.yml#L16) configuration.

`SERVICE_PORT: 80` -> `SERVICE_PORT: 8080`

Vitals9367:
Changed service port to 8080

nikomakela (approved):
looks good to me, but I have no experience with these

nikomakela:
Is there a good reason why the builds fails?

Jokinen:
The build likely fails because of permissions. The account of the person opening the PR has to have enough privileges, e.g. be a member of City of Helsinki, or some group within it.

PR #411, opened 2022-03-22
KK-840 | As a child, if I have already participated in 2 events, I can still see new events in UI but cannot enrol to them

Jokinen:
## Description

* Children are now limited to two enrolments per year (configured per project)
* Guardians can see events and event groups their child is already enrolled in
* Layout changes
* Additional copy to guide the user

Design can be found from here: https://app.abstract.com/projects/447ded2e-2949-4a2b-a67e-874ab853cf1d/branches/master/collections/b6d80365-e04d-40e0-865e-02021a61a80c

## Context

[KK-840](https://helsinkisolutionoffice.atlassian.net/browse/KK-840)

Screenshot 2022-03-22 at 16 38 47
Screenshot 2022-03-22 at 16 39 40
Screenshot 2022-03-22 at 16 39 40 1
Screenshot 2022-03-22 at 16 39 48
Screenshot 2022-03-22 at 16 39 58
Screenshot 2022-03-22 at 16 40 06


src/domain/child/ChildEnrolmentCount.tsx
```diff
@@ -0,0 +1,52 @@
...
+            <LoadingSpinner
+              theme={{
+                '--spinner-color': 'var(--color-black)',
+              }}
+              small
+            />
+          )}
+        </>
+      }
+    />
```
nikomakela, line 50: Quite hard to read... 🤔

nikomakela (commented):
I find this new design difficult to read, because there is too much text without clear containers: 
image. More spacing between the child profile sections?

This loading spinner spins forever without any visible error messages:
image

On next page load, it was the another spinner that spinned forever
image

src/domain/child/ChildEnrolmentCount.tsx
nikomakela: could the loading spinner and the actual text content be in variables? Then they could be assigned to the text with a string-formatter `${t('child.message.eventVisitsThisYear')} ${data ? counter ? loadingspinner}` and there would be no need for the empty fragment either.

nikomakela (approved):
I would try some other layout designs, because I find this page heavy really to read. The code changes looks good! Lets collect other opinions from test server :)

src/domain/child/ChildEnrolmentCount.tsx
nikomakela: The spinners started to work better later on. Some problems with infra. The pill component could look less like a button when there would be a colon: "Tapahtumakäynnit tänä vuonna: ..."

Jokinen: I don't mind making this change. I'd perhaps personally prefer your version, but it often gets flagged as unclear as well. Some devs find it difficult to understand because you have to "jump around" on the file and you can't just browse it from top to bottom.

I added the colon.

Jokinen: Unfortunately we can't use string templates/concatenation here as the spinner's as the "primitive" type a component maps to is a function that can't be stringified by default. I implemented your suggestion to the degree it was possible.

Jokinen:
Rebased on top of current main to resolve conflicts.

PR #413, opened 2022-03-29
KK-873, KK-876 | Render menu and pages from CMS

Jokinen:
## Description

* Navigation component is now rendered with a HDS component
* Pages from Kukkuu's Headless CMS are now available at their `path`
* Changes to language
   * Don't use redirects to ensure a language exists. Instead consider a missing language parameter to mean Finnish.
   * Sync language change with `i18n.changeLanguage` instead of a hard refresh
   * When i18next detects user language, enforce that detected language by adding it to the pathname
   * All previous paths should still exist

## Context

[KK-873](https://helsinkisolutionoffice.atlassian.net/browse/KK-873)
[KK-876](https://helsinkisolutionoffice.atlassian.net/browse/KK-876)


nikomakela (approved):
Code changes looks good! Some bugs though:

### Language change

1.  If the cms page does not exist on selected language, the result page is pure white page without any errors. Instead of that it should give a proper explanation what happened and the page has finished the loading. Steps: Go to a CMS page that exists in some language, but not in every language, then change the language to one that does not have the page yet. E.g. http://localhost:3000/en/page-1/ ->  http://localhost:3000/sv/page-1/. Of course the user can also go directly to the (missing) URL.

### Links

1. There is a page that contains a link to the Google. It's an external link, but it does not pop up in a new tab. It does not have an indicator or accessibility label that tells that it's an external link. (EDIT: I was told that the target can be set in CMS-admin, but the indicator is still missing)

browser-tests/childrenFeature.ts
```diff
@@ -86,16 +86,10 @@ test('As a guardian I want to edit the details of my child', async (t) => {
...
-    postalCode,
-    firstName,
-    lastName,
-    relationship,
-  } = t.ctx.addChild;
+  const { birthDate, city, postalCode, firstName, lastName, relationship } =
+    t.ctx.addChild;
   const newChildName = `${firstName} ${lastName}`;
-  const newChildNameRegepx = new RegExp(`${newChildName} .*`);
+  const newChildNameRegepx = new RegExp(newChildName);
```
badrange, line 92: There is a speling error in this const, it should be newChildNameRegexp and not ..epx

Jokinen: Long time no see 🙂 Thanks for spotting this one, fixed 👍

badrange: Long time indeed! Interesting to follow the development of this project from afar!

Jokinen:
I've added fixes to the `react-helsinki-headless-cms` library that address these issues.

Missing translations are now detected as missing pages. I assumed that a page's slug was language specific, but this isn't the case.

I've adjusted the behaviour of links within content. The `rhhc` library now provides a flavour of HDS' `Link` component that takes data from the CMS and detects corrects props to pass to HDS' component.

When a content editor links a page, the CMS still build `a` elements that that have `href` values that point to the API. I've asked about this in the headless-cms channel.

I think that we should avoid mutating the data we get from the CMS so I've not added any hackfixes for now. After we get an answer we'll know better.

In the meantime content editors can build links by manually linking to pages.

PR #417, opened 2022-03-31
Fixed the back-button path on Ticketmaster enrolment page

nikomakela:
KK-921

### DIFF
--- a/src/domain/event/EventOccurrenceRedirect.tsx
+++ b/src/domain/event/EventOccurrenceRedirect.tsx
@@ -2,7 +2,7 @@ import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import copy from 'copy-to-clipboard';
 import { IconCheck } from 'hds-react';
-import { useLocation, useParams } from 'react-router';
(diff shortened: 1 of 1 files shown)


src/domain/event/EventOccurrenceRedirect.tsx
```diff
@@ -16,6 +16,7 @@ import LoadingSpinner from '../../common/components/spinner/LoadingSpinner';
 import InfoPageLayout from '../app/layout/InfoPageLayout';
 import useEventOccurrence from './queries/useEventOccurrence';
 import useAriaLive from '../../common/AriaLive/useAriaLive';
+import useGetPathname from '../../common/route/utils/useGetPathname';
 
 function getEventPathname(pathname: string): string {
```
Jokinen, line 21: Unused function

Jokinen: The `useLocation` hook is also no longer used.

For some reason GH doesn't allow me to comment on that row directly.

Jokinen (approved):

PR #419, opened 2022-04-04
KK-875 | Show notifications from the CMS

Jokinen:
## Description

Display notifications from the CMS on all pages.

Notification can be dismissed. A notification dismissal is remembered over sessions.

Integrates the `Notification` component from the RHHC library.

Contains removal of some dead code that I noticed while removing the unused variable `$headerHeight`.

## Context

[KK-875](https://helsinkisolutionoffice.atlassian.net/browse/KK-875)

## Screenshots

Screenshot 2022-04-04 at 11 40 44
Screenshot 2022-04-04 at 11 40 50


nikomakela (approved):
Good!

PR #422, opened 2022-04-05
KK-920 | Fix comments made by Anna

Jokinen:
## Description

* Don't show new event label if child can't enrol into upcoming events
* Include service description text on child page
* Fix modal alignment

## Context

[KK-920](https://helsinkisolutionoffice.atlassian.net/browse/KK-920)

### DIFF
--- a/src/common/translation/i18n/fi.json
+++ b/src/common/translation/i18n/fi.json
@@ -148,6 +148,91 @@
     },
     "skipToContentLabel": "Siirry suoraan sisältöön"
   },
+  "profile": {
+    "common": {
+      "message": {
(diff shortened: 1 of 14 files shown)


src/common/translation/i18n/fi.json
```diff
@@ -148,6 +148,91 @@
     },
     "skipToContentLabel": "Siirry suoraan sisältöön"
   },
+  "profile": {
+    "common": {
+      "message": {
+        "serviceDescription": "Hienoa että olet mukana Kulttuurin kummilapset -toiminnassa! Täällä voit tarkastella tapahtumakutsuja ja ilmoittautua tapahtumiin. Huomioithan, että kummilapsi voi osallistua vuosittain kahteen tapahtumaan, ellei toisin ilmoiteta. Nähdään kulttuurin parissa!"
```
charn, line 154: Should this be "Hienoa että" -> "Hienoa, että"

Jokinen: Hmn, maybe 🤔 Can't recall these rules in enough detail. This is how the copy is in the design.

Jokinen: Perhaps "Hienoa, että", unravels into: "On hienoa, että" which would suggest that the comma should be there.

PR #426, opened 2022-04-06
Remove unnecessary colon

Jokinen:
## Description

Translations contained a colon that was included by accident, resulting in a double colon in the UI. Remove colon from translations.

Other

- Removes package-lock file that should not be in the repo. We use yarn.lock instead.


nikomakela (approved):

nikomakela:
Approved! But, I wouldn't do a hotfix-release myself because of this

Jokinen:
Dropped documentation for release.

PR #429, opened 2022-04-25
Fix login when using some other language than Finnish

Jokinen:
## Description

Previously logging in would fail when users used some other language than Finnish.

When the user logged in while using English:

1) They get redirected from Tunnistamo to `/callback#...`
2) i18next detects that user locale is `en` from a cookie
3) i18next language is changed
4) `languageChanged` hook is fired in `src/common/translation/i18n/i18nInit.ts`
5) User is redirected into `/callback` (missing hash)

### DIFF
--- a/src/common/route/useSyncLanguageBetweenI18nAndReactRouter.ts
+++ b/src/common/route/useSyncLanguageBetweenI18nAndReactRouter.ts
@@ -6,7 +6,7 @@ import { SUPPORT_LANGUAGES } from '../../common/translation/TranslationConstants
 import { localeParam } from '../../domain/app/appRoutes';
 
 function persistLanguageChoice(language: string) {
-  document.cookie = `i18next=${language}; SameSite=Strict`;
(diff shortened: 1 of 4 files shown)


nikomakela (approved):
I'm not sure if it was just an issue in review server or on my machine, but it was really hard to logout. After logout, the only screen is to log me out from the github. If I used the backbutton to go back to Kukkuu, Tunnistamo always aggressively tried to login, so I couldn't get to the home page (or the index) of the Kukkuu. Also the language did not change properly if only the backbutton was used. The language had to be changed once with a page reload  or full page load before it was set properly.

Anyway, this works better with the earlier tests.

Jokinen:
Hmn, I was only able to reproduce these issues intermittently. I'll see if I can quickly pin down a cause.

Jokinen:
I wasn't able to pinpoint the cause for the login loop.

I found a bug in language persistence that I resolved.

Could you test again to see if this language changing bug was the root cause of the redirection bug? If not, would you happen to remember the steps you took that caused the bug?

PR #436, opened 2022-08-24
HH-152: update rhhc and change link props + updated/refactored default configuration

nenonja:
## Description
Updated RHHC and link component props api
https://helsinkisolutionoffice.atlassian.net/browse/HH-152
+ also refactored the configuration to use useRHHCConfig with the latest configuration from RHHC
+ added new variable to all the environemnts, so needed to change github actions as well
+ also added the deployment-url to the pull-request for easier testing

Changes are testable here: https://kukkuu-ui-hh-152-refactor-link-component-436.test.kuva.hel.ninja/
## Context


nikomakela (changes requested):
The changes that are made looks OK, but the updated HCRC package requires more configurations

```
Property 'internalHrefOrigins' is missing in type '{ siteName: string; apolloClient: ApolloClient<NormalizedCacheObject>; currentLanguageCode: any; components: { Img: (props: React.ImgHTMLAttributes<HTMLImageElement>) => JSX.Element; A: ({ href, ...delegatedProps }: React.AnchorHTMLAttributes<...>) => JSX.Element; Link: ({ href, target, ...delegatedProps }: React.An...' but required in type 'Config'
```

nenonja:
> The changes that are made looks OK, but the updated HCRC package requires more configurations
> 
> ```
> Property 'internalHrefOrigins' is missing in type '{ siteName: string; apolloClient: ApolloClient<NormalizedCacheObject>; currentLanguageCode: any; components: { Img: (props: React.ImgHTMLAttributes<HTMLImageElement>) => JSX.Element; A: ({ href, ...delegatedProps }: React.AnchorHTMLAttributes<...>) => JSX.Element; Link: ({ href, target, ...delegatedProps }: React.An...' but required in type 'Config'
> ```

Fixed now by updating the configuration and also refactored the configuration to use new useRHHCConfig hook same way as in hobbies. I find it a little bit problematic that I have to fix somebody else's changes on RHHC, there is kind of techical debt. I would prefer that if one changes the RHHC he has to also update all the projects to use the latest verson and check they still work. If we move to monorepo someday, this would be automatically done.

nenonja:
I added the new environment variable to the review, staging and production ymls, so perhaps @terovirtanen can check those parts?

terovirtanen (approved):
Github workflow files are ok

nenonja:
Browser tests are failing but they seem to fail also on master. Here is another pr just to compare against the master and tests fail there too: https://github.com/City-of-Helsinki/kukkuu-ui/pull/438

package.json
```diff
@@ -44,6 +44,7 @@
   "dependencies": {
     "@apollo/client": "3.4.16",
     "@datapunt/matomo-tracker-react": "0.5.1",
+    "@ffmpeg-installer/ffmpeg": "^1.1.0",
```
nenonja, line 47: dev package needed for recording testcafe videos on review environment

nikomakela:
> > The changes that are made looks OK, but the updated HCRC package requires more configurations
> > ```
> > Property 'internalHrefOrigins' is missing in type '{ siteName: string; apolloClient: ApolloClient<NormalizedCacheObject>; currentLanguageCode: any; components: { Img: (props: React.ImgHTMLAttributes<HTMLImageElement>) => JSX.Element; A: ({ href, ...delegatedProps }: React.AnchorHTMLAttributes<...>) => JSX.Element; Link: ({ href, target, ...delegatedProps }: React.An...' but required in type 'Config'
> > ```
> 
> Fixed now by updating the configuration and also refactored the configuration to use new useRHHCConfig hook same way as in hobbies. I find it a little bit problematic that I have to fix somebody else's changes on RHHC, there is kind of techical debt. I would prefer that if one changes the RHHC he has to also update all the projects to use the latest verson and check they still work. If we move to monorepo someday, this would be automatically done.

It is problematic indeed, but the RHHC is a library that should be used by every app of the Solution office and the library itself should be maintained somehow externally. We can never get in situation where we would update every application after every RCHC-lib change, because we don't have time for that and we don't have budget for that. The latest changes in RCHC were actually not even needed by the Kukkuu, because so far Kukkuu has not shown any interest to carousels or article archives or any other new features that the newer version of the RCHC has. Even if we are at the moment having a 1 single team that maintains all the apps that are using the RCHC, that won't be the case in the future.

nikomakela:
There are 2 CMS pages in the test cms server:

image

When the page is opened, the HCRC-lib always renders a grey koros to the bottom of the page - I find out that it was an empty list of collection components. It looked like there were 2 footers and the other one of them was just broken:

image

This is actually a bug in the HCRC and not in the Kukkuu. I can make a change for that. But, there is a workaround that does not need updates from HCRC:

```
        <PageContent
          notFoundPageContent={<NotFound />}
          collections={() => null}
        />
```

nikomakela:
Fixed: 
image

nikomakela (approved):
After the commit that I already pushed, I think the PR is good. If the browser test is misbehaving, should we just disable it? Otherwise it gives "false positives" and the real issues are missed.

nikomakela:
Btw, the conventional commits are not in use in this project... just FYI, if you want your commits to look similar with others :)

nenonja:
Ok, great! I will merge this now

PR #439, opened 2022-09-06
Browser tests with github login

terovirtanen:
## Description

- Browser tests with github login
  - github login is used when testcafe has option '--dev'
- Command 'test:browser' is meant for developer usage and it uses github login by default
- New user registration for browser tests

## Context

[KK-954](https://helsinkisolutionoffice.atlassian.net/browse/KK-954)


package.json
```diff
@@ -23,8 +23,9 @@
     "update-translations": "fetch-translations 1b5qAamjhmSNiME3matBINaeyz8TvoRWVB6wU5HvKy4g -l en,fi,sv -o src/common/translation/i18n",
     "test": "TZ=UTC react-scripts test --transformIgnorePatterns \"node_modules/(?!(react-helsinki-headless-cms|react-helsinki-headless-cms/apollo))/\"",
     "test:debug": "react-scripts --inspect-brk test --runInBand --no-cache",
-    "test:browser": "testcafe \"chrome --window-size='1920,1080'\" browser-tests/ --live --lang=fi-FI",
-    "test:browser:ci": "testcafe \"chrome:headless --disable-gpu --window-size='1920,1080'\" browser-tests/ -s takeOnFails=true --lang=fi-FI"
+    "test:browser": "testcafe \"chrome --window-size='1920,1080'\" browser-tests/ --live --dev --lang=fi-FI --skip-js-errors",
```
nenonja, line 26: why `--skip-js-errors`? Sometimes errors are important information. 
If you use something like this https://github.com/City-of-Helsinki/yjdh/blob/develop/frontend/testcafeClientErrorHandler.js
You can use that to filter out specific known errors that are not important like ResizeObserver

browser-tests/pages/register.ts
```diff
@@ -0,0 +1,34 @@
+import { screen } from '@testing-library/testcafe';
+import { Selector } from 'testcafe';
+
+export const registerForm = {
```
nenonja, line 4: instead of simple const object, I would rather use class based pagemodel which is recommended by testcafe: https://github.com/DevExpress/testcafe-examples/blob/master/examples/use-page-model/page-model.js
You can also add actions there as methods like clicking single button or even longer operation like filling form and submit.

browser-tests/utils/login.ts
```diff
@@ -10,13 +11,33 @@ const givePermission = async (t: TestController) => {
 };
 
 export const login = async (t: TestController) => {
-  await t
-    .click(ssoLogin.loginLink)
-    .typeText(ssoLogin.username, testUsername())
-    .typeText(ssoLogin.password, testUserPassword())
-    .click(ssoLogin.loginButton);
+  // @ts-ignore
+  const developmentMode = t.testRun.opts.developmentMode
```
nenonja, line 15: `--dev` is not a bad idea but I also wonder if just to use another `.env` flag variable like `GITHUB_LOGIN=True` 
or `BROWSER_TESTS_LOCAL = True`. Then you can even run the browser tests against staging or production by changing env parameters. Now you have to have also remove the `--dev` from the yarn script if you want to run tests against some other environment.

nenonja (approved):
Looks good! Some minor comments, but still approved. PS: I merged my pr and it caused a conflict, so you have to resolve that first :)

package.json
nenonja: You have to refer to that handler from testcafe settings: https://github.com/City-of-Helsinki/yjdh/blob/1abe148649a7699941ce262b094ac36a5c61e3ff/frontend/.testcaferc.base.js#L19

terovirtanen: I don't know is this real problem (and how to fix it) or should this just skip in browser tests.
I created a bug report  about this error
https://helsinkisolutionoffice.atlassian.net/browse/KK-955

browser-tests/utils/login.ts
terovirtanen: I think '--dev' -option do not affect to tests itself or limit used environments. 
This is used to check if tests are run by user (use github login) or CI automation (use helsinki tunnus login). 
  
testcafe documentation:
https://testcafe.io/documentation/402638/reference/configuration-file#developmentmode

_**developmentMode**_
_Enables mechanisms to log and diagnose errors. Enable this option if you plan to contact TestCafe Support to report an issue._

package.json
terovirtanen: Only this specific error note is skipped, details are documented on KK-955

browser-tests/utils/login.ts
terovirtanen: Keep as it is now

browser-tests/pages/register.ts
terovirtanen: Keep this as it is now, class based pagemodel will use with new tests

PR #441, opened 2022-09-12
KK-898 | Show also Ticketmaster enrolment in child's events page

tuomas777:
## Description

Added Ticketmaster based event enrolments to child's page. This required switching to the backend's new query
`child.activeInternalAndTicketmasterEnrolments` from the old `child.occurrences` query, which returned only internal event enrolments. The new query returns enrolments instead of occurrences, so quite a lot of non-visible changes were needed.

Ticketmaster enrolments are shown a bit differently than internal enrolments. They will contain the event's image instead of the QR code because we don't know the code, and they don't contain occurrence info, but instead a mention about the event's registration being handled in Ticketmaster. There is already more stuff planned to be added to Ticketmaster enrolments, at least a button to mark the enrolment attended, but that will be implemented separately later.

Event sections' heading texts on the child events page were also fine-tuned a bit to better describe the current situation.

## Context

[KK-898](https://helsinkisolutionoffice.atlassian.net/browse/KK-898)

## Screenshots

Screenshot 2022-09-12 at 9 54 28


nikomakela (commented):
Overall, quite ok, but I would make some bigger refactoring in the side of these changes. Since there were quite a few of them, I created my own PR #444 where I have done the changes that I'm suggesting. The PR is basically done to improve the typing and taking advantage of them.

Also, with the improved typing and code refactoring, some issues appeared. For example, I think it was actually a lucky coincidence that the function that handles both the event types did not raise errors -- the method was done in the return / render, which I think is too late for type some type checks.

nikomakela (approved):
Approvable after changes from #444

PR #445, opened 2022-09-20
KK-899 | Add enrolled Ticketmaster event page

tuomas777:
Added a new event page for Ticketmaster events that users have "enrolled" (obtained a password) in. One will end up on this new page when clicking the event in the upcoming events section of a child's profile page.

The new page contains the same kind of Ticketmaster password showing functionality as exists already on a different page, so the password thing was refactored into its own separate component.

Screenshot 2022-09-23 at 15 45 10

### DIFF
--- a/src/common/translation/i18n/en.json
+++ b/src/common/translation/i18n/en.json
@@ -344,6 +344,12 @@
     },
     "ticketmasterNotice": "Registration for the event takes place in the Ticketmaster ticket service. To reserve tickets you will need a user account which you can create during the reservation process if you do not already have one. The up to date status for placing can be found from Ticketmaster."
   },
+  "ticketmasterEvent": {
+    "description": "",
+    "passwordLabel": "",
+    "continueButton": "",
+    "participantsPerInviteText": ""
+  },
   "eventOccurrenceRedirectPage": {
     "back": "Back",
     "continue": "Continue to Ticketmaster",
--- a/src/domain/event/TicketmasterEventIsEnrolled.tsx
+++ b/src/domain/event/TicketmasterEventIsEnrolled.tsx
@@ -1,0 +1,75 @@
+import { useTranslation } from 'react-i18next';
+import { useParams } from 'react-router-dom';
+import { useQuery } from '@apollo/client';
+
+import styles from './ticketmasterEventIsEnrolled.module.scss';
+import { ticketmasterEventQuery } from './queries/ticketmasterEventQuery';
+import {
+  ticketmasterEventQuery as TicketmasterEventQueryType,
+  // eslint-disable-next-line max-len
+  ticketmasterEventQuery_event_ticketSystem_TicketmasterEventTicketSystem as EventTicketSystem,
+  // eslint-disable-next-line max-len
+  ticketmasterEventQuery_event_occurrences_edges_node_ticketSystem_TicketmasterOccurrenceTicketSystem as OccurrenceTicketSystem,
+} from '../api/generatedTypes/ticketmasterEventQuery';
+import LoadingSpinner from '../../common/components/spinner/LoadingSpinner';
+import Paragraph from '../../common/components/paragraph/Paragraph';
+import EventPage from './EventPage';
+import ErrorMessage from '../../common/components/error/Error';
+import { useChildRouteGoBackTo } from '../profile/route/ProfileRoute';
+import AnchorButton from '../../common/components/button/AnchorButton';
+import TicketmasterPassword from './TicketmasterPassword';
+import Text from '../../common/components/text/Text';
+
+const TicketmasterEventIsEnrolled = () => {
+  const { t } = useTranslation();
+  const goBackTo = useChildRouteGoBackTo();
+  const params = useParams<{ childId: string; eventId: string }>();
+  const { loading, error, data } = useQuery<TicketmasterEventQueryType>(
+    ticketmasterEventQuery,
+    {
+      variables: {
+        eventId: params.eventId,
+        childId: params.childId,
+      },
+    }
+  );
+
+  if (loading) return <LoadingSpinner isLoading={true} />;
+
+  const errorMessage = <ErrorMessage message={t('api.errorMessage')} />;
+  if (error || !data?.event) return errorMessage;
+
+  const eventTicketSystem = data?.event.ticketSystem as EventTicketSystem;
+  /* TODO Currently we don't have Ticketmaster URLs for whole events, so as a workaround
+      the URL of the first occurrence is used for now.
+   */
+  const ticketmasterUrl = (
+    data?.event?.occurrences?.edges[0]?.node
+      ?.ticketSystem as OccurrenceTicketSystem
+  ).url;
+
+  return (
+    <EventPage event={data.event} backTo={goBackTo}>
+      <div className={styles.description}>
+        <Paragraph text={data.event.description || ''} />
+      </div>
+      <Text>{t('ticketmasterEvent.description')}</Text>
+      <Text>{t('ticketmasterEvent.passwordLabel')}</Text>
+      <div className={styles.passwordRow}>
+        <TicketmasterPassword password={eventTicketSystem?.childPassword} />
+        <AnchorButton href={ticketmasterUrl} openInNewTab>
+          {t('ticketmasterEvent.continueButton')}
+        </AnchorButton>
+      </div>
+      <Text>
+        {t('ticketmasterEvent.participantsPerInviteText', {
+          participantsPerInvite: t(
+            `event.participantsPerInviteEnum.${data.event.participantsPerInvite}`
+          ),
+        })}
+      </Text>
+    </EventPage>
+  );
+};
+
+export default TicketmasterEventIsEnrolled;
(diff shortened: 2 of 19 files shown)


src/domain/event/TicketmasterEventIsEnrolled.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const ticketmasterUrl = (
+    data?.event?.occurrences?.edges[0]?.node
+      ?.ticketSystem as OccurrenceTicketSystem
+  ).url;
+
+  return (
+    <EventPage event={data.event} success={<SuccessToast />} backTo={goBackTo}>
+      <div className={styles.description}>
+        <Paragraph text={data.event.description || ''} />
+      </div>
```
nikomakela, line 51: trivial thing, but this could be without div, because it's unnecessary and other paragraphs don't have a div either.

src/domain/event/__tests__/__snapshots__/TicketmasterEventIsEnrolled.test.tsx.snap
```diff
@@ -0,0 +1,14 @@
...
+exports[`renders snapshot correctly 1`] = `
+<div>
+  <div
+    aria-label="Lataa"
+    class="spinnerWrapper"
+  >
+    <div
+      class="spinner"
+    />
+  </div>
```
nikomakela, line 12: Spinner is not something we want here :)

src/domain/event/queries/ticketmasterEventQuery.ts
```diff
@@ -0,0 +1,32 @@
...
+export const ticketmasterEventQuery = gql`
+  query ticketmasterEventQuery($eventId: ID!, $childId: ID!) {
+    event(id: $eventId) {
+      id
+      name
+      description
+      image
+      imageAltText
+      participantsPerInvite
+      occurrences: occurrences(upcoming: true, first: 1) {
```
nikomakela, line 12: are they sorted properly?

src/domain/event/__tests__/__snapshots__/TicketmasterEventIsEnrolled.test.tsx.snap
tuomas777: Hah. That test was actually copied straight from another already existing test, and didn't check it out properly yet. The old test has that spinner snapshot too, so I guess it is just some kind of smoke test or something. Anyway, seems somewhat pointless, so removed that new test.

src/domain/event/queries/ticketmasterEventQuery.ts
tuomas777: Should be, but that doesn't actually even really matter, because this is just a temporary workaround until we get just one URL for whole events.

src/domain/event/TicketmasterEventIsEnrolled.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <Text>{t('ticketmasterEvent.description')}</Text>
+      <Text>{t('ticketmasterEvent.passwordLabel')}</Text>
+      <div className={styles.passwordRow}>
+        <TicketmasterPassword password={eventTicketSystem?.childPassword} />
+        <AnchorButton href={ticketmasterUrl}>
```
nikomakela, line 61: This should take the user to a new tab. Also it should tell about that for a screen reader.

nikomakela: I'm not sure, but probably this can take `target="_black"` and `aria-label={ t('common.openInNewTabAriaLabel')}` as an attribute.

nikomakela (changes requested):
The button that takes the user to external page should always open in a new tab. Also the accessibility of this feature should be taken care with an area-label that says "Opens in a new tab".

image

A translation is missing

image

There were also some English and Swedish translations still missing.

src/domain/event/TicketmasterEventIsEnrolled.tsx
nikomakela: I think there should also be an "external link" -icon in it.

src/domain/event/__tests__/__snapshots__/TicketmasterEventIsEnrolled.test.tsx.snap
nikomakela: it could also just wait that the spinner is gone!

src/common/translation/i18n/en.json
```diff
@@ -344,6 +344,12 @@
     },
     "ticketmasterNotice": "Registration for the event takes place in the Ticketmaster ticket service. To reserve tickets you will need a user account which you can create during the reservation process if you do not already have one. The up to date status for placing can be found from Ticketmaster."
   },
+  "ticketmasterEvent": {
+    "description": "",
+    "passwordLabel": "",
+    "continueButton": "",
+    "participantsPerInviteText": ""
+  },
```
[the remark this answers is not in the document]
tuomas777: Yeah of course those are needed before this goes live, but they aren't available yet. Personally I prefer leaving missing translations empty rather than filing them with some half-baked DIY versions I will then forget to update 😄

src/domain/event/TicketmasterEventIsEnrolled.tsx
tuomas777: Not sure about this, in [HDS principles](https://hds.hel.fi/components/link#principles) it says _Generally, links should not open in a new tab_.

src/domain/event/TicketmasterPassword.tsx
```diff
@@ -0,0 +1,58 @@
...
+      // to copy by other means.
+      if (success) {
+        sendMessage(t('eventOccurrenceRedirectPage.passwordCopySuccess'));
+        setCopyStatus(CopyStates.success);
+      }
+    }
+  };
+
+  return (
+    <>
```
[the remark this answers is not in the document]
tuomas777: Problem with a div is that creating the needed layouts becomes more cumbersome.

src/domain/event/ticketmasterPassword.module.scss
```diff
@@ -0,0 +1,40 @@
...
+  color: $color-black-90;
+
+  background: none;
+  border: none;
+  cursor: pointer;
+}
+
+.copyButtonWrapper {
+  position: relative;
+  width: fit-content !important;
```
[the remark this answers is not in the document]
tuomas777: Tweaked the styles to get rid of those.

src/domain/event/TicketmasterPassword.tsx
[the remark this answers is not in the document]
tuomas777: Changed to div.

src/domain/event/TicketmasterEventIsEnrolled.tsx
nikomakela: It also says that 

> HDS Link supports a link with an external icon. Use the external prop when you wish to navigate to an entirely different site.

and (https://hds.hel.fi/components/link/accessibility) 

> Remember to tell users if the link is going to open an entirely new site. In HDS React Link this feature is provided through external links (prop external). There is a default aria-label for screen readers. This can be overwritten by using the openInExternalDomainAriaLabel prop. In HDS Core, set a proper aria-label to the anchor tag as shown in the examples below.

and 

> Remember to tell users if the link is going to open in a new tab. In HDS React Link this feature is provided through prop openInNewTab. There is a default aria-label for screen readers. This can be overwritten by using the openInNewTabAriaLabel prop. In HDS Core, set a proper aria-label to the anchor tag as shown in the examples below.

Also we get some rejecting comments from auditors when an accessibility audit was made to Tapahtumat.

nikomakela (commented):
The changes looks good. The external link missing still
- [x] accessibility of the external link (to ticketmaster)

src/domain/event/TicketmasterEventIsEnrolled.tsx
tuomas777: Added the external link icon and changed the thing to open in a new tab.

src/domain/event/TicketmasterEventIsEnrolled.tsx
```diff
@@ -0,0 +1,74 @@
...
+        <TicketmasterPassword password={eventTicketSystem?.childPassword} />
+        <AnchorButton href={ticketmasterUrl} openInNewTab>
+          {t('ticketmasterEvent.continueButton')}
+        </AnchorButton>
+      </div>
+      <Text>
+        {t('ticketmasterEvent.participantsPerInviteText')}
+        {t(
+          `event.participantsPerInviteEnum.${data.event.participantsPerInvite}`
+        ) + '.'}
```
nikomakela, line 68: I would include the dot in the translation, but it's ok this time.

nikomakela (approved):
The button text is not included in the aria-label, so the voice over says only "Opens in new tab", but does not tell where it navigates.
image

image

When the button text is included in the aria label so that screen reader can read it, this PR is ready to be shipped. Approved already.

src/domain/event/TicketmasterEventIsEnrolled.tsx
tuomas777: Yeah it is a bit meh, but that same text is already used in another place in the UI where it should not contain a dot, so we would need to maintain the same texts with and without the dot, which is IMO a worse option.

PR #458, opened 2023-02-22
Fix registration page redirect when profile exists

nikomakela:
KK-989. Redirect to the profile page when registrating with an existing credentials.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+# 1.10.1
+
+### Fixed
+
+- Fix registration page redirect when profile exists
+
 # 1.10.0
 
 ### Added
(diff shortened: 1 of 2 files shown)


nikomakela:
> In the ticket it was wanted that the user should be redirected to the "Add child" form directly, now the user is redirected to the profile page but the "Add child" form remains unopened. I tried opening the "Add child" form after having been redirected to the profile page and it did indeed open up with the correct birthdate and home municipality of the child to be added. Is the ticket's description up to date? If it is then the automatic opening of the "Add child" form needs to be added still IMHO.

hmm.. okay. I'll ask from the product owner. In my opinion, you would more often just click a wrong button when you want to login than you would add new kids there. If I would add a new kid to any service, I would think that I would first need to go to my profile page or similar. The redirect to the profile page was also something that was actually already done earlier inside a component, but that feature was accidentally removed, when the whole route to the component was removed. So it has earlier worked like this too.

nikomakela:
review url: https://kukkuu-ui-kk-989-fix-logged-in-register-v-c3.test.kuva.hel.ninja/en/

jannetalvio:
Works as expected.

jannetalvio (approved):
Works as expected.

PR #460, opened 2023-03-21
Fix the URL locale handling

nikomakela:
KK-995.

The locale was replaced from anywhere from the URL pathname,
when it should have been replaced only from the beginning.
This led to a situation where e.g `/enrol` part of the pathname
was replaced with `rol`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+# 1.10.2
+
+### FIXED
+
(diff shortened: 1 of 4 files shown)


karisal-anders (approved):

karisal-anders:
@nikomakela Although I approved already I still want to point out the possible edge cases where there is no language code in the URL's start but something else which has the language code as a prefix e.g. /finlands-opening-ceremony/ or /ennen-jotain/. These could still be mangled by `replaceLocaleInPathname`.

PR #492, opened 2023-09-27
Upgrade all the dependencies and migrate the react-router

nikomakela:
KK-1017. 

1. Start using the Node 20.
2. Upgrade all the deps with `yarn-upgrade-all`. 
3. Migrate to React-router v6. A migration guide: https://reactrouter.com/en/main/upgrading/v5.
4. Remove the Enzyme which was used only to test snapshots of the components and replace them with snapshots done by testing library.
5. Migrate API token service methods to be static.
6. Remove finally-polyfill.
7. Fix the Moment lib timezone handling after dep updates
8. Fix the navigation confirm prompt
9. Refactor the language / locale handling: Since the router update, the pathname should always have the locale set (except in auth callback).
10. Refactor the Headless CMS route.
11. Migrate to router provider- Use `<RouterProvider>` component such that it would enable the new Data APIs on routes defined at the root router. More about: https://reactrouter.com/en/6.16.0/upgrading/v6-data.

----

NOTE: The moment lib should be replaced next. I prefer date-fns.

PR #493, opened 2023-09-27
Migrate to Vite

nikomakela:
Bases to the huge pile of upgrading and migration changes in https://github.com/City-of-Helsinki/kukkuu-ui/pull/492.

The Create React App (CRA) is deprecated. It is advised to replace the CRA with Vite.

The migration process is done by following the instructions from
https://www.robinwieruch.de/vite-create-react-app/:
- The index.html is moved from the `/public` to the root. The relative paths are migrated.
-  Vitest is installed and the configuration for testing-library is made

The Sass styles from HDS and the app itself needed a couple of aliases.

The new setup includes
- Vite instead of CRA
- Vitest with the Testing-library instead of Jest
- Typescript for Vite

More about this:
- https://blog.bitsrc.io/the-future-of-react-why-create-react-app-is-deprecated-and-hooks-are-the-future-83e8a087a325
- https://dev.to/ag2byte/create-react-app-is-officially-dead-h7o
- https://github.com/facebook/create-react-app/issues/13072


nikomakela:
Actually, right after merging this to #492 , I realized that the pipeline is not ready for Vite's environment variables, so I reverted branch deletion. Unfortunately this PR is still shown as merged and I cannot remerge it, but the plan was to first take the #492 through the pipeline and then configure the pipeline to use vite-configs.

nikomakela:
This PR is recreated here: https://github.com/City-of-Helsinki/kukkuu-ui/pull/496

PR #550, opened 2024-04-24
Use authorization code flow and HDS login handler

nikomakela:
KK-1097 KK-1126.

NOTE: At this point, the (local) Tunnistamo response type is changed
from implicit flow to the authorization code flow.

- Install the latest  version of the HDS-react.
- Use the login provider that the HDS offers.
- Get rid of the old OIDC implementation

refactor: remove needless code from the auth folder

refactor: remove the CRA service worker
The app is no longer CRA app, but a Vite app.

refactor: update react-helmet-async

refactor: remove the old auth state for redux

fix: route titles usage

fix: logout profile state handling and use api token from hds login provider

refactor: user loading in profile page

refactor: deprecate profile reducer in favor of guardian provider

refactor: add refetch profile

fix: profile uses the useProfile at right time

fix: the error with helmet while login process

refactor: use a text callback handler to prevent flickering

fix: login process redirect to profile page

feat: unauthorized will redirect to profile if user has logged in

fix: follow the hook rules with HOCs

refactor: navigate to unauthorized when no permission to AppRoute

fix: profileChildDetail rendering with login nextPath

fix: remove react strict mode for better experience with hds login

fix: login issues in child route

chore: upgrade vitest

refactor: configurations for both server types

refactor: change all the react-router imports to react-router-dom

refactor: upgrade the react-helsinki-headless-cms -plugin

feat: audiences can be configured with env variables

The default OIDC server should be set to Tunnistamo, while the Keycloak
version of the Kukkuu UI cannot be published yet.

chore: add comment about React.Strict and HANDLING_LOGIN_CALLBACK

The HDS Login component is triggering HANDLING_LOGIN_CALLBACK during the
login process, since it is not working well with the React.Strict mode.

refactor: add oidc return type as configurable

refactor: remove UI reducer and redux-oidc


terovirtanen:
## TestCafe result is __success__ for https://kukkuu-local.api.dev.hel.ninja :laughing::tada::tada::tada:

terovirtanen:
## TestCafe result is __failed__ for https://kukkuu-ui-pr550.dev.hel.ninja :crying_cat_face::anger::boom::boom:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-local.api.dev.hel.ninja :laughing::tada::tada::tada:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-local.api.dev.hel.ninja :laughing::tada::tada::tada:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-local.api.dev.hel.ninja :laughing::tada::tada::tada:

terovirtanen:
**KUKKUU-UI branch is deployed to platta: https://kukkuu-ui-pr550.dev.hel.ninja** :rocket::rocket::rocket:

package.json
```diff
@@ -102,14 +102,14 @@
     "prop-types": "^15.8.1",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
-    "react-helmet-async": "^1.3.0",
+    "react-helmet-async": "^2.0.4",
     "react-helsinki-headless-cms": "1.0.0-alpha253",
```
karisal-anders, line 106: If you're upgrading hds-react to v3.7.0 I think it'd be good to upgrade the react-helsinki-headless-cms to the newest one too as then they'd be using the same version of hds-react.

nikomakela: Good call. Upgraded

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-local.api.dev.hel.ninja :laughing::tada::tada::tada:

terovirtanen:
**KUKKUU-UI branch is deployed to platta: https://kukkuu-ui-pr550.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __failed__ for https://kukkuu-ui-pr550.dev.hel.ninja :crying_cat_face::anger::boom::boom:

src/domain/auth/constants.ts
```diff
@@ -0,0 +1,51 @@
...
+  ...providerPropertiesBase,
+  apiTokensClientSettings: {
+    url: `${AppConfig.oidcAuthority}protocol/openid-connect/token`,
+    maxRetries: 10,
+    retryInterval: 1000,
+    queryProps: {
+      grantType: 'urn:ietf:params:oauth:grant-type:uma-ticket',
+      permission: '#access',
+    },
+    audiences: ['kukkuu-api-test', 'profile-api-test'],
```
nikomakela, line 49: I think this should be configurable with the environment variables. If local-dev-kukkuu-ui wants to interact with the local-dev-kukkuu-api, I think this audience is not right. I think it needs to be changed to `kukkuu-api-dev` in these situations.

- [ ] Make the audience configurable.

src/domain/api/client.ts
```diff
@@ -40,26 +40,12 @@ const errorLink = onError(({ graphQLErrors, networkError }) => {
...
-        );
-      }
-
       if (errorCode === 'AUTHENTICATION_ERROR') {
         // It is not possible to recover from AUTHENTICATION_ERROR at least with the
         // same API token, so to minimize further problems it is probably best to just
         // log the user out completely.
-        logoutTunnistamo();
+        removeApiTokensFromStorage();
+        removeUserReferenceFromStorage();
```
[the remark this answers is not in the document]
nikomakela: No, it is actually the right and the same thing. Removing the user and api tokens from the storage is what the logout should do, but there is no logout command available (without a hook). Instead calling these 1 by 1, there seems to be a call that clears all the storages https://github.com/City-of-Helsinki/helsinki-design-system/blob/5b2e9b243f76c5a1d685ba9d49e35f036310bc07/packages/react/src/components/login/apiTokensClient/apiTokensClient.ts#L216. Maybe that could be called instead.

src/domain/app/AppConfig.ts
```diff
@@ -11,15 +12,38 @@ class AppConfig {
...
+  }
+
+  static get oidcKukkuuApiClientId() {
+    return getEnvOrError(
+      import.meta.env.VITE_OIDC_KUKKUU_API_CLIENT_ID,
+      'VITE_OIDC_KUKKUU_API_CLIENT_ID'
+    );
+  }
+
+  static get oidcServerType(): 'KEYCLOAK' | 'TUNNISTAMO' {
```
karisal-anders, line 45: Adding type checking:
```typescript
  static get oidcServerType(): OidcServerType {
    const oidcServerType = import.meta.env.VITE_OIDC_SERVER_TYPE ?? 'KEYCLOAK';
    if (!OIDC_SERVER_TYPES.includes(oidcServerType)) {
      throw new Error(`Invalid OIDC server type: ${oidcServerType}`);
    }
    return oidcServerType;
  }
```
with
```typescript
const OIDC_SERVER_TYPES = ['KEYCLOAK', 'TUNNISTAMO'] as const;
type OidcServerType = (typeof OIDC_SERVER_TYPES)[number];
```
somewhere.

src/domain/api/client.ts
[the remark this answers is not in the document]
nikomakela: If you are on some restricted page, the page component itself should do the redirecting to right route. What the actual logout-command does (the oidc client logout), it seems to jsut emit a signal for logout and nothing listens that signal (in HDS). Then after signal emitting, it calls OIDC usermanage logour redirect url.

src/domain/app/AppConfig.ts
karisal-anders: But it's a bit hard to determine if the types really work as the "yarn typecheck" doesn't say anything about the AppConfig.ts file. I changed the code something that definitely should've broken the types and still "yarn typecheck" passed fine.

src/domain/app/routes/SilentRenewRedirect.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { useNavigate } from 'react-router-dom';
+
+const SilentRenewRedirect = () => {
+  const navigate = useNavigate();
+  React.useEffect(() => {
+    navigate('/silent_renew.html', { replace: true });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
karisal-anders, line 9: Why is the eslint-disable-next-line necessary? Now the useEffect runs only once, right? But if the `navigate` function was added to the dependency array then the useEffect would be run each time the `navigate` function changes? But when does it change? I tried looking into react-router's useNavigate function (react-router-dom imports useNavigate from react-router according to its source code in node_modules) and `node_modules/react-router/dist/index.js:
```typescript
function useNavigate() {
  let {
    isDataRoute
  } = React.useContext(RouteContext);
  // Conditional usage is OK here because the usage of a data router is static
  // eslint-disable-next-line react-hooks/rules-of-hooks
  return isDataRoute ? useNavigateStable() : useNavigateUnstable();
}
```

So it's a bit, hmm, interesting, let's say. So if `isDataRoute` in `RouteContext` is truish then a stable version of the navigate function is returned:
```typescript
/**
 * Stable version of useNavigate that is used when we are in the context of
 * a RouterProvider.
 */
function useNavigateStable() {
...
```
but otherwise an unstable version.

How do you see this? Did you encounter problems if you added the `navigate` function to the dependency array?

src/domain/app/routes/WithLocaleRoute.tsx
```diff
@@ -2,16 +2,19 @@ import React from 'react';
 
 import { useLocaleRouteNavigate } from './useLocaleRouteNavigate';
 
+const WithLocalRouteComponent = ({ Component, ...rest }: any) => {
+  useLocaleRouteNavigate();
+  return <Component {...rest} />;
+};
```
karisal-anders, line 8: Local vs. Locale here. What is meant here?

src/domain/app/routes/WithLocaleRoute.tsx
```diff
@@ -2,16 +2,19 @@ import React from 'react';
...
  * the router will navigate to such a route and chagne the location pathname.
  */
 const WithLocalRoute = (WrappedComponent: React.FC) => {
   return ({ ...props }) => {
-    // eslint-disable-next-line react-hooks/rules-of-hooks
-    useLocaleRouteNavigate();
-    return <WrappedComponent {...props} />;
+    return <WithLocalRouteComponent Component={WrappedComponent} {...props} />;
   };
 };
```
karisal-anders, line 19: Reading the comment/documentation it seems `Locale` is meant here instead of `Local`, right?

src/domain/app/AppRoute.tsx
```diff
@@ -25,30 +22,38 @@ function AppRoute({
...
+
   return (
     <>
       {/* eslint-disable-next-line @typescript-eslint/ban-ts-comment */}
       {/* @ts-ignore ts2786 - https://github.com/nfl/react-helmet/issues/646 */}
       <Helmet>
         {title && (
           <title>
-            {title} - {t('appName')}
+            {t(title)} - {t('appName')}
```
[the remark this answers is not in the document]
karisal-anders: Renaming the `title` to `titleKey` and adding `const title = t(titleKey)` right after useTranslation call in tandem with the added documentation to AppRouteProps might be even clearer, and it would handle the truish checks for `title` like it was before.

src/domain/app/routes/SilentRenewRedirect.tsx
nikomakela: I would self use it totally without the useEffect, but a warning was printed to console, that the navigate should always be called from a useEffect-callback. When the dependency array is empty, it really is ran only once, but the redirect should be needed only once. I would say that the way how it's currently used should be fine and it would probably do the same if the the navigate would be in the useEffect, (but I would like to call it only once).

src/domain/app/routes/WithLocaleRoute.tsx
nikomakela: Right, it's a typo. Needs top be fixed.

src/domain/auth/KukkuuHDSLoginCallbackhandler.tsx
karisal-anders, line 1: A small thing but the filename ends with `handler` and the function is `Handler` i.e. case difference.

src/domain/app/AppRoute.tsx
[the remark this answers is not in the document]
nikomakela: The earlier implementation was that the translation was already done in the router, but it didn't do anything 😞 The translation was needed to be done later on.

I will change the property name.

src/domain/auth/KukkuuHDSLoginCallbackhandler.tsx
```diff
@@ -0,0 +1,49 @@
...
+    if (!error) return;
+    if (
+      error.type === 'SIGNIN_ERROR' &&
+      error.message ===
+        'Current state (HANDLING_LOGIN_CALLBACK) cannot be handled by a callback'
+    ) {
+      // TODO: resolve this issue -- why is it raised?
+      // We don't want to handle this error,
+      // because it is raised during every login process,
+      // and it does not affect any how to the signin process.
```
karisal-anders, line 34: https://github.com/City-of-Helsinki/helsinki-design-system/blob/v3.7.0/packages/react/src/components/login/components/LoginCallbackHandler.tsx#L39

&

https://github.com/City-of-Helsinki/helsinki-design-system/blob/v3.7.0/packages/react/src/components/login/client/oidcClient.ts#L255-L276

are at least related...

src/domain/auth/README.md
```diff
@@ -0,0 +1,52 @@
...
+}
+```
+
+# Keycloak (test environment)
+
+```json
+{
+  "userManagerSettings": {
+    "authority": "https://tunnistus.test.hel.ninja/auth/realms/helsinki-tunnistus/",
+    "client_id": "kukkuu-ui-dev",
```
karisal-anders, line 33: This (i.e. client_id="kukkuu-ui-dev") is for review/development environments according to the [Palveluintegraatio document](https://helsinkisolutionoffice.atlassian.net/wiki/spaces/KAN/pages/8679030785/Palveluintegraatio+HP+-+Kukkuu+Kulttuurin+kummilapset), not test environment. So maybe switching to "kukkuu-ui-test" or renaming "test environment" would be clearer.

src/domain/api/client.ts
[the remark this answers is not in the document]
karisal-anders: Hmm... so `logoutFromOidc` in `useLogout` hook does this:
```typescript
    logout();
    removeApiTokensFromStorage();
    removeUserReferenceFromStorage();
    clearProfile();
    flushAllState({ keepUserFormData: false });
```

And ´logout()` and `clearProfile()` need hooks to work... but `flushAllState({keepUserFormData: false})` should work without hooks. Is there a case where calling `flushAllState` would do something useful here?

Maybe moving the non-hook stuff into one function, and calling that function in `useLogout` (+the hook needing stuff) and here would make it harder for the logout functionality that is needed to go out of sync between different places in the code. What do you think?

src/domain/profile/ProfileProvider.tsx
```diff
@@ -0,0 +1,53 @@
...
+  });
+
+  const isLoading = loading && !called;
+
+  const clearProfile = () => setProfileToContext(null);
+
+  const refetchProfile = () => {
+    (async () => await refetch())();
+    setProfileToContext(data?.myProfile || null);
+  };
```
karisal-anders, line 29: Hmm... doesn't it now work so that if you call `refetchProfile` it runs the first line with the function call `(async () => await refetch())();` but as the function is asynchronous it won't wait for the function call to return anything but will just immediately run the following line `setProfileToContext(data?.myProfile || null);`? Is your intention here to wait for the refetching to happen first and after that has happened, then to set the refetched myProfile data into the context?

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-local.api.dev.hel.ninja :laughing::tada::tada::tada:

src/domain/app/AppConfig.ts
karisal-anders: Well, as you are typing the return value to be either `KEYCLOAK` or `TUNNISTAMO` without checking that they actually are then the type system is fooled here into showing stuff that isn't really necessarily true. So either removing the return typing (worse IMO) or making the necessary checks in order to really to be able to type the return value as either `KEYCLOAK` or `TUNNISTAMO` would be good. And if "yarn typecheck" is not working for AppConfig.ts then it would be better to fix that so it does work.

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-local.api.dev.hel.ninja :laughing::tada::tada::tada:

src/domain/auth/README.md
karisal-anders: Isn't "kukkuu-ui-test" the client ID used for test environment? At least according to the [Palveluintegraatio document](https://helsinkisolutionoffice.atlassian.net/wiki/spaces/KAN/pages/8679030785/Palveluintegraatio+HP+-+Kukkuu+Kulttuurin+kummilapset).

src/domain/profile/ProfileProvider.tsx
nikomakela: The refetch is an async function, so it needs to be waited, because it returns a promise. The refetchProfile however, is not an async function and I don't even want it to be one. Using a then-method to attach the callback function could look better here, but as far as I know, the result should be exaxctly the same.

I can change this to 
```
    refetch().then(() => {
      setProfileToContext(data?.myProfile || null);
    });
```

karisal-anders (commented):
Just went though all the changes so far, took over 4 hours. Left some comments, please have a look at them. We could have a chat tomorrow, e.g. with huddle if needed still, live conversation could be useful as this is quite a complex PR.

terovirtanen:
**KUKKUU-UI branch is deployed to platta: https://kukkuu-ui-pr550.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __failed__ for https://kukkuu-ui-pr550.dev.hel.ninja :crying_cat_face::anger::boom::boom:

src/domain/profile/modal/EditMyProfileForm.tsx
```diff
@@ -77,18 +79,34 @@ export default function EditMyProfileForm({
...
+            ...(prevValue ?? initialValues),
+            ...input,
+            languagesSpokenAtHome: {
+              __typename: 'LanguageNodeConnection',
+              edges:
+                input.languagesSpokenAtHome?.map((id) => ({ node: { id } })) ??
+                [],
+            },
+          }) as MyProfile
+      );
```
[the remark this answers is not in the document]
nikomakela: It sure would be better, but I need to double check whether it is OK for the API, because then also the updateMyProfile-mutation would use those default values.

src/domain/app/AppConfig.ts
karisal-anders: I think the version we came up together in a huddle is ok now.

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-local.api.dev.hel.ninja :laughing::tada::tada::tada:

src/domain/profile/ProfileProvider.tsx
nikomakela: An IIFE (Immediately Invoked Function Expression) is a JavaScript function that runs as soon as it is defined:
https://developer.mozilla.org/en-US/docs/Glossary/IIFE

nikomakela: It really seems that it is not waiting for the response
https://stackoverflow.com/a/54758455.

nikomakela: So, the then-block should be better and it's now in use. Good call!

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-local.api.dev.hel.ninja :laughing::tada::tada::tada:

src/domain/auth/README.md
nikomakela: Yes, but this is a local machine configuration when the local kukkuu-ui is connected to the test-environment. So, a new paragraph really is needed. :) 

The kukkuu-ui-dev is the only one that has the localhost-addresses set in the callback urls. The test-profile is configured so that the dev-client can use it.

I have already tested this configuration that it works: this is the actual configuration that I was running the local client.

src/domain/auth/KukkuuHDSLoginCallbackhandler.tsx
nikomakela: I actually do know what the reason for this is. It's a bug in the HDS-react package. It is not tested with so big application as the Kukkuu UI is: The Kukkuu UI has so many providers / container that makes the login provider to rerender, that it then raises those errenous signals that it should not raise. They promised to fix it.
```
      // The HANDLING_LOGIN_CALLBACK error is raised only
      // when we are using the <React.Strict>.
      // We don't want to handle this error,
      // because it is raised during every login process,
      // and it does not affect any how to the signin process.
      // When the <BrowserApp> is removed from the React.Strict -container,
      // The error won't be raised anymore. The HDS team has been noted about this issue.
```

src/domain/registration/form/RegistrationForm.tsx
```diff
@@ -177,6 +200,20 @@ const RegistrationForm = () => {
...
+                          ...prevStatus,
+                          ...response.data?.submitChildrenAndGuardian?.guardian,
+                          languagesSpokenAtHome: {
+                            edges: values.guardian.languagesSpokenAtHome.map(
+                              (language) => ({
+                                node: { id: language, language },
+                              })
+                            ),
+                          },
+                        }) as MyProfile
```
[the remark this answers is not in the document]
nikomakela: This is the clearest solution I could come up with at this point. I would like to get rid of the whole redux-storage, so we could revisit this then.

src/domain/profile/modal/EditMyProfileForm.tsx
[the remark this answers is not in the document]
nikomakela: I made some changes which based on this and the guard is now on.

src/domain/api/client.ts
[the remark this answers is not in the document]
nikomakela: Yes, I can add the `flushAllState`. Good!

terovirtanen:
**KUKKUU-UI branch is deployed to platta: https://kukkuu-ui-pr550.dev.hel.ninja** :rocket::rocket::rocket:

src/domain/auth/README.md
nikomakela: It's a shame that we cannot see the actual configuration anyhow -- we always need to trust in 2nd party and in the SERvice integration document, that does not actualyl tell the configuration related to the client ids, scopes nor the callback urls.

terovirtanen:
## TestCafe result is __failed__ for https://kukkuu-ui-pr550.dev.hel.ninja :crying_cat_face::anger::boom::boom:

src/domain/api/client.ts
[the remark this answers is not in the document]
nikomakela: I think the clearProfile should actually be called automatically as a side effect by the signals that the remove commands send. The clearProfile is a context method so it is in the render stack and so cannot be accessed outside a React component, but the provider/component should rerender when a signal is sent, so the profile should be set to null.

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-local.api.dev.hel.ninja :laughing::tada::tada::tada:

terovirtanen:
**KUKKUU-UI branch is deployed to platta: https://kukkuu-ui-pr550.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __failed__ for https://kukkuu-ui-pr550.dev.hel.ninja :crying_cat_face::anger::boom::boom:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-local.api.dev.hel.ninja :laughing::tada::tada::tada:

terovirtanen:
**KUKKUU-UI branch is deployed to platta: https://kukkuu-ui-pr550.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __failed__ for https://kukkuu-ui-pr550.dev.hel.ninja :crying_cat_face::anger::boom::boom:

src/domain/profile/ProfileProvider.tsx
nikomakela: This describes this better https://stackoverflow.com/questions/58027037/unexpected-behavior-async-await-when-required-with-iife

src/domain/auth/README.md
nikomakela: Changed

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-local.api.dev.hel.ninja :laughing::tada::tada::tada:

terovirtanen:
**KUKKUU-UI branch is deployed to platta: https://kukkuu-ui-pr550.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __failed__ for https://kukkuu-ui-pr550.dev.hel.ninja :crying_cat_face::anger::boom::boom:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-local.api.dev.hel.ninja :laughing::tada::tada::tada:

terovirtanen:
**KUKKUU-UI branch is deployed to platta: https://kukkuu-ui-pr550.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __failed__ for https://kukkuu-ui-pr550.dev.hel.ninja :crying_cat_face::anger::boom::boom:

src/domain/app/routes/WithLocaleRoute.tsx
```diff
@@ -2,16 +2,19 @@ import React from 'react';
...
+  useLocaleRouteNavigate();
+  return <Component {...rest} />;
+};
+
 /**
  * Check that the current pathname starts with a current locale.
  * If the current locale is not set in the pathname as the first segment,
  * the router will navigate to such a route and chagne the location pathname.
  */
 const WithLocalRoute = (WrappedComponent: React.FC) => {
```
karisal-anders, line 15: The `Local` vs. `Locale` typo still here

src/domain/app/AppRoute.tsx
```diff
@@ -1,54 +1,60 @@
...
     // eslint-disable-next-line no-console
     console.warn(
       // eslint-disable-next-line max-len
-      'Most app routes should have a title. If a route does not need one, explicitly state is by toggling the noTitle prop.'
+      'Most app routes should have a title. If a route does not need one, explicitly state is by toggling the noTitle prop.',
+      { element }
     );
   }
 
+  const title = titleKey ? `${t(titleKey)} - ${t('appName')}` : t('appName');
```
karisal-anders, line 42: You can leave this as it is if you like, but moving the `const title = ...` right after the `const { t } = useTranslation();` and then using the `title` instead of `titleKey` after that in the code it'd be more like it was before and it would also check whether the `titleKey` resolves to a non-empty string or not.

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-local.api.dev.hel.ninja :laughing::tada::tada::tada:

terovirtanen:
**KUKKUU-UI branch is deployed to platta: https://kukkuu-ui-pr550.dev.hel.ninja** :rocket::rocket::rocket:

karisal-anders (approved):
Great job tackling this beast of a change! 👍 Looks ok to merge!

terovirtanen:
## TestCafe result is __failed__ for https://kukkuu-ui-pr550.dev.hel.ninja :crying_cat_face::anger::boom::boom:

nikomakela:
> Great job tackling this beast of a change! 👍 Looks ok to merge!

I've been investigating whether this new login provider has a backward support for the old implicit flow of Tunnistamo implementation that is currently active in all the public envs and so far it seems like there is something that breaks the bakward compatibility. 😞 If it really is so, it means that we cannot merge this before we really want to start using the Keycloak, which would mean that we should change all the pipelines and make it sure that the Keycloak is ready for the production use. I've contacted the the HDS team about this.

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-local.api.dev.hel.ninja :laughing::tada::tada::tada:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-local.api.dev.hel.ninja :laughing::tada::tada::tada:

terovirtanen:
## TestCafe result is __failed__ for https://kukkuu-ui-pr550.dev.hel.ninja :crying_cat_face::anger::boom::boom:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-local.api.dev.hel.ninja :laughing::tada::tada::tada:

terovirtanen:
**KUKKUU-UI branch is deployed to platta: https://kukkuu-ui-pr550.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __failed__ for https://kukkuu-ui-pr550.dev.hel.ninja :crying_cat_face::anger::boom::boom:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-local.api.dev.hel.ninja :laughing::tada::tada::tada:

nikomakela:
> > Great job tackling this beast of a change! 👍 Looks ok to merge!
> 
> I've been investigating whether this new login provider has a backward support for the old implicit flow of Tunnistamo implementation that is currently active in all the public envs and so far it seems like there is something that breaks the bakward compatibility. 😞 If it really is so, it means that we cannot merge this before we really want to start using the Keycloak, which would mean that we should change all the pipelines and make it sure that the Keycloak is ready for the production use. I've contacted the the HDS team about this.
> 
> When the implicit flow and the old configurations are used, the new login provider does nothing visible and it does not forward the user to the Tunnistamo. The following error is thrown:
> 
> ```
> [UserManager] signinRedirect: begin
> HomeHero.tsx:60 [RedirectNavigator] prepare: begin
> chunk-TOGC4UFP.js?v=5bb69a6b:28932 [UserManager] _signinStart: begin
> chunk-TOGC4UFP.js?v=5bb69a6b:28932 [OidcClient] createSigninRequest: begin
> chunk-TOGC4UFP.js?v=5bb69a6b:28932 [UserManager] _signinStart: error after preparing navigator, closing navigator window
> chunk-TOGC4UFP.js?v=5bb69a6b:28932 [RedirectNavigator] close: begin
> ```
> 
> I revalidated my settings by using a master branch and there the configurations worked, so there is something breaking in the new login provider.
> 
> If the oidc return type is set to `VITE_OIDC_RETURN_TYPE="code"`, then the UI reacts, but is not working, because Tunnistamo is set in implicit flow mode. When the implicit flow return type is used with `VITE_OIDC_RETURN_TYPE="id_token token"`, it does not do anything but throws an error to console when a debug mode is on.

It seems like the implicit flow is totally removed from the oidc-client-ts: https://github.com/authts/oidc-client-ts/issues/152. That plugin is used as a base for HDS Login. However, I tested that we can simply activate another OIDC return type on side of another, so we can have the both active: `id_token token` (the implicit flow) and the `code` (authorization code flow).

When the authorization code flow is activated in the Tunnistamo (no other configurations are needed), the new implementation should work OK.

> NOTE: The current test (and staging?) environment is using the Kuva-instance of the Tunnistamo.

terovirtanen:
**KUKKUU-UI branch is deployed to platta: https://kukkuu-ui-pr550.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**KUKKUU-UI branch is deployed to platta: https://kukkuu-ui-pr550.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-local.api.dev.hel.ninja :laughing::tada::tada::tada:

terovirtanen:
## TestCafe result is __failed__ for https://kukkuu-ui-pr550.dev.hel.ninja :crying_cat_face::anger::boom::boom:

nikomakela:
The new `VITE_OIDC_KUKKUU_API_CLIENT_ID="https://api.hel.fi/auth/kukkuu"` env variable needs to be added to the pipelines in order to get the current Tunnistamo implementation to work with the new HDS login provider.

PR #597, opened 2024-10-22
fix: upgrade apollo client to fix navigation issues

nikomakela:
KK-1248.

Fixes the issue when a user navigated from the event page to the event redirect page while enrolling to an event of an external ticket system.

----
Related: https://github.com/apollographql/apollo-client/issues/11639

The issue was:
> Console has a following error “Uncaught TypeError: Cannot convert object to primitive value“, but it does not tell where it happens.
Steps to reproduce:
Click “Jatka ilmoittautumaan“
Result: The loading spinner spins forever and an error is written in the console.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "set-value": ">=4.0.1"
   },
   "dependencies": {
-    "@apollo/client": "^3.7.16",
+    "@apollo/client": "^3.10.1",
     "@faker-js/faker": "^8.1.0",
     "@ffmpeg-installer/ffmpeg": "^1.1.0",
     "@jonkoops/matomo-tracker-react": "^0.7.0",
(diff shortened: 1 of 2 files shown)


karisal-anders (approved):
Good! Knowing this I think it's a relatively safe bet that there are probably problems in other apps as well using RHHC and some other version of apollo-client which is not compatible with ^3.10.1, so it would be good to take a look at them as well.

terovirtanen:
**KUKKUU-UI branch is deployed to platta: https://kukkuu-ui-pr597.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __failed__ for https://kukkuu-ui-pr597.dev.hel.ninja :crying_cat_face::anger::boom::boom:

PR #607, opened 2024-11-22
fix: use prod variable

melniiv:
## Context


terovirtanen:
## TestCafe result is __success__ for https://kukkuu-ui-pr607.dev.hel.ninja :laughing::tada::tada::tada:

src/index.tsx
```diff
@@ -12,10 +12,12 @@ import AppConfig from './domain/app/AppConfig';
...
 Modal.setAppElement('#root');
 
-if (import.meta.env.VITE_ENVIRONMENT === 'production') {
+console.warn('Environment prod: ', import.meta.env.PROD);
+
+if (import.meta.env.PROD) {
   Sentry.init({
     dsn: import.meta.env.VITE_SENTRY_DSN,
-    environment: import.meta.env.VITE_ENVIRONMENT,
+    environment: import.meta.env.PROD ? 'production' : 'development',
```
nikomakela, line 18: **This change should be rejected!** Environment was set properly: It would be clear in the Sentry that it was the review-server or the test server that sent the error. No need to change this! This `import.meta.env.VITE_ENVIRONMENT` is set to "review" in development pipelines, meaning the PR-servers. It's set to testing, staging and production in the other pipelines that we have.

src/index.tsx
```diff
@@ -12,10 +12,12 @@ import AppConfig from './domain/app/AppConfig';
 initI18next();
 Modal.setAppElement('#root');
 
-if (import.meta.env.VITE_ENVIRONMENT === 'production') {
+console.warn('Environment prod: ', import.meta.env.PROD);
+
+if (import.meta.env.PROD) {
```
nikomakela, line 15: This is the important change! 👍 

Now it's right, since the `import.meta.env.PROD` in VITE is same as `process.env.NODE_ENV === 'production'` in many other Node frameworks. The `import.meta.env.PROD` should be `true` everytime the app package is built and bundled (like the NODE_ENV === 'production' in similar conditions). 

The Sentry SDK should not be initialized in development environment (meaning developers local computers). There is no option to disable the Sentry after it's initialized.

References:
- https://vite.dev/guide/env-and-mode#node-env-and-modes
- https://v2.vitejs.dev/guide/env-and-mode.html#env-variables-and-modes

src/domain/app/BrowserApp.tsx
```diff
@@ -33,7 +33,7 @@ const matomoInstance = createMatomoInstance({
 // Prevent non-production data from being submitted to Matomo
 // by pretending to require consent to process analytics data and never ask for it.
 // https://developer.matomo.org/guides/tracking-javascript-guide#step-1-require-consent
-if (import.meta.env.VITE_ENVIRONMENT !== 'production' && window?._paq) {
+if (!import.meta.env.PROD && window?._paq) {
```
nikomakela, line 36: Now when I see the actual use cases, I think the VITE_ENVIRONMENT actually really had a reason to be there. Do we want the matomo to be in use in test and stage environments? Because this will make them available there. `import.meta.env.VITE_ENVIRONMENT !== 'production'` was explciitly for the production environment, because VITE_ENVIRONMENT is set to 'production' only in the prod env (kummilapset.hel.fi).

Dockerfile
```diff
@@ -54,7 +54,6 @@ ARG VITE_ORIGIN
 ARG VITE_ADMIN_TICKET_VALIDATION_URL
 ARG VITE_API_URI
 ARG VITE_CMS_URI
-ARG VITE_ENVIRONMENT
```
nikomakela, line 57: We can keep this to indicate the real actual environment. The usage: https://dev.azure.com/City-of-Helsinki/kukkuu/_search?action=contents&text=VITE_ENVIRONMENT&type=code&lp=code-Project&filters=ProjectFilters%7Bkukkuu%7DRepositoryFilters%7Bkukkuu-pipelines%7D&pageSize=25&result=DefaultCollection/kukkuu/kukkuu-pipelines/GBmaster//devops/kukkuu-ui-build-config.yml

nikomakela (changes requested):
The VITE_ENVIRONMENT seem to have had a real purpose, so I'm starting to think it should not be removed. Only the Sentry was initialization was done wrong earlier, but now it's fixed, which is a good thing.

src/index.tsx
nikomakela: the environment option in Sentry init configuration is to define a request parameter when the log is created in sentry. It defines the environment (review, testing, staging, production).

The Sentry should be initialized only when a built & bundled version is running (`import.meta.env.PROD === true` or `NODE_ENV === 'production'`)

src/domain/app/BrowserApp.tsx
nikomakela: We have in AppConfig 

```typescript
  static get matomoEnabled() {
    return Boolean(
      getEnvOrError(import.meta.env.VITE_MATOMO_ENABLED, 'VITE_MATOMO_ENABLED')
    );
  }
 ```

**Comment by terovirtanen** [comment]
**KUKKUU-UI branch is deployed to platta: https://kukkuu-ui-pr607.dev.hel.ninja** :rocket::rocket::rocket:

**Comment by terovirtanen** [comment]
## TestCafe result is __success__ for https://kukkuu-ui-pr607.dev.hel.ninja :laughing::tada::tada::tada:

**Comment by nikomakela** [comment]
Should we use the AppConfig for these? It seems to be hard to understand what the environment variables can have as a value and what is their use case. If we would have getters in the AppConfig, we could describe it better. 

```typescript
/**
 * Allowed environments for the application.
 */
const ALLOWED_ENVIRONMENTS = ['development', 'review', 'testing', 'staging', 'production'] as const;

/**
 * Type alias for allowed environments.
 */
type Environment = typeof ALLOWED_ENVIRONMENTS[number];

/**
 * Typeguard to check if a string is a valid `Environment`.
 *
 * @param env - The string to check.
 * @returns `true` if the string is a valid `Environment`, `false` otherwise.
 */
function isEnvironment(env: string): env is Environment {
  return ALLOWED_ENVIRONMENTS.includes(env as Environment);
}

class AppConfig {
  /**
   * Checks if the application is running in production mode.
   *
   * Whether the app is running in production (running the dev server with NODE_ENV='production' or running an app built with NODE_ENV='production').
   *
   * @returns `true` if the app is in production mode, `false` otherwise.
   */
  static get isAppInProductionMode(): boolean {
    return import.meta.env.PROD;
  }

  /**
   * Gets the current environment of the application.
   *
   * The possible values are defined by `ALLOWED_ENVIRONMENTS`.
   * Defaults to 'development' if the environment variable `VITE_ENVIRONMENT` is not set.
   *
   * @returns The current environment as a string.
   * @throws {Error} If the environment variable `VITE_ENVIRONMENT` has an invalid value.
   */
  static get environment(): Environment {
    const env = (import.meta.env.VITE_ENVIRONMENT as Environment) || 'development';

    if (!isEnvironment(env)) {
      throw new Error(`Invalid environment: ${env}`);
    }

    return env;
  }
}
```

src/domain/headlessCms/client.ts
```diff
@@ -21,7 +22,7 @@ const errorLink = onError(({ graphQLErrors, networkError }) => {
 
       Sentry.captureMessage(errorMessage);
 
-      if (import.meta.env.NODE_ENV === 'development') {
+      if (AppConfig.environment === 'development') {
```
nikomakela, line 25: This also works, but I think we would want it to reflect on whether the code is in bundled prod mode:
```suggestion
      if (!AppConfig.isAppInProductionMode) {
```

Dockerfile
nikomakela: This VITE_ENVIRONMENT is now in use in AppConfig, so it is needed here in build args too.

nikomakela (changes requested):
The `VITE_ENVIRONMENT` is needed in the build args. Also, I would change the 1 if statement.

src/domain/headlessCms/client.ts
nikomakela: The 2 main options of the old value `import.meta.env.NODE_ENV` are "development" and "production". That is what we can get with `AppConfig.isAppInProductionMode`. There could have also been NODE_ENV == "test", but that is as well something that we would have liked to have in unit testing. That's why I think using `isAppInProductionMode` is a better option.

nikomakela (approved):

terovirtanen:
**KUKKUU-UI branch is deployed to platta: https://kukkuu-ui-pr607.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-ui-pr607.dev.hel.ninja :laughing::tada::tada::tada:

PR #641, opened 2025-01-20
chore: docker compose production port forward

nikomakela:
KK-1405


karisal-anders (approved):
Any particular reason for using port 8082 rather than e.g. 8081 if kukkuu-admin would use 8082? Now you're using 8082 for kukkuu-ui and 8084 for [kukkuu-admin](https://github.com/City-of-Helsinki/kukkuu-admin/pull/309).

PR #662, opened 2025-02-21
feat: use apollo client local storage cache persister

nikomakela:
KK-1406.

Installed apollo3-cache-persist.

Refactored code by spitting the CMS client code to multiple files.

Implemented `TimedApolloCachePersistor` that creates a new persisted Apollo cache persistor with an expiration time. This class automatically clears expired caches and provides methods for persisting and restoring the cache to/from localStorage.

The language(s) queries are now hard coded to the Apollo client's in memory cache in the apollo client's initialization, so there is no need to handle cache or queries related those anymore.

### DIFF
--- a/src/domain/headlessCms/client.ts
+++ b/src/domain/headlessCms/client.ts
@@ -1,52 +1,103 @@
-import {
-  ApolloClient,
-  createHttpLink,
-  InMemoryCache,
-  ApolloLink,
-} from '@apollo/client';
+import { ApolloClient, createHttpLink, ApolloLink } from '@apollo/client';
 import { onError } from '@apollo/client/link/error';
 import * as Sentry from '@sentry/browser';
-import { LanguagesDocument } from 'react-helsinki-headless-cms/apollo';
 
 import AppConfig from '../app/AppConfig';
-import { HARDCODED_CMS_LANGUAGE_QUERY_RESPONSE } from '../languages/constants';
+import {
+  createApolloCache,
+  isConfiguredQueryPolicy,
+  queryPolicies,
+} from './cache';
+import { LogLevel, TimedApolloCachePersistor } from './persistor';
 
-const httpLink = createHttpLink({
-  uri: import.meta.env.VITE_CMS_URI,
-});
+export function createCmsApolloClient() {
+  const httpLink = createHttpLink({
+    uri: import.meta.env.VITE_CMS_URI,
+  });
 
-const errorLink = onError(({ graphQLErrors, networkError }) => {
-  if (graphQLErrors) {
-    graphQLErrors.forEach(({ message, locations, path, extensions }) => {
-      const errorCode = extensions?.code;
+  const errorLink = onError(({ graphQLErrors, networkError }) => {
+    if (graphQLErrors) {
+      graphQLErrors.forEach(({ message, locations, path, extensions }) => {
+        const errorCode = extensions?.code;
 
-      // eslint-disable-next-line max-len
-      const errorMessage = `[GraphQL error]: Message: ${message}, Code: ${errorCode}, Location: ${locations}, Path: ${path}`;
+        // eslint-disable-next-line max-len
+        const errorMessage = `[GraphQL error]: Message: ${message}, Code: ${errorCode}, Location: ${locations}, Path: ${path}`;
 
-      Sentry.captureMessage(errorMessage);
+        Sentry.captureMessage(errorMessage);
 
-      if (!AppConfig.isAppInProductionMode) {
-        // eslint-disable-next-line no-console
-        console.error(errorMessage);
-      }
-    });
-  }
+        if (!AppConfig.isAppInProductionMode) {
+          // eslint-disable-next-line no-console
+          console.error(errorMessage);
+        }
+      });
+    }
 
-  if (networkError) {
-    Sentry.captureMessage('Network error');
-  }
-});
+    if (networkError) {
+      Sentry.captureMessage('Network error');
+    }
+  });
 
-const client = new ApolloClient({
-  link: ApolloLink.from([errorLink, httpLink]),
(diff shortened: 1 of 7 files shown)


src/domain/headlessCms/persistor.ts
```diff
@@ -0,0 +1,203 @@
...
+    }
+  }
+}
+
+export interface TimedApolloCachePersistorOptions {
+  /**
+   * The time-to-live for the persisted cache in milliseconds.
+   * Defaults to 10 minutes (600000 milliseconds).
+   */
+  persistedCacheTimeToLive?: number;
```
karisal-anders, line 60: How about `persistedCacheTimeToLiveMs` or `persistedCacheTimeMsToLive`? It'd be clear at a glance for the user that this is milliseconds, which would be good for clarity IMHO.

src/domain/headlessCms/persistor.ts
```diff
@@ -0,0 +1,203 @@
...
+   * @returns The configured CachePersistor instance.
+   */
+  private createApolloCachePersistor(cache: InMemoryCache) {
+    if (this.hasPersistedCacheExpired()) {
+      try {
+        this.clearLocalStorage();
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      } catch (e: any) {
+        this.logger.error('Failed to clear local storage:', e.message, e.stack);
+      }
```
karisal-anders, line 118: How about something like this?
```suggestion
      } catch (e: unknown) {
        const errorMessage = 'Failed to clear local storage:';
        if (e instanceof Error) {
          this.logger.error(errorMessage, e.message, e.stack);
        } else {
          this.logger.error(errorMessage, e);
        }
      }
```
To get rid of the eslint disabling, and still handling `Error` type ok, and having a fallback for other raised values.

src/domain/headlessCms/client.ts
```diff
@@ -37,16 +36,63 @@ const errorLink = onError(({ graphQLErrors, networkError }) => {
   }
 });
 
-const client = new ApolloClient({
-  link: ApolloLink.from([errorLink, httpLink]),
-  cache: new InMemoryCache(),
+const cache = createApolloCache();
+const kukkuuApolloCachePersistor = new TimedApolloCachePersistor(cache, {
+  logLevel: LogLevel.WARN,
```
nikomakela, line 41: - [ ] Add time to live time by using AppConfig.

src/domain/headlessCms/persistor.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  /**
+   * Checks if the persisted Apollo cache has expired based on the
+   * `persistedCacheTimeToLive` value.
+   *
+   * @returns `true` if the cache has expired, `false` otherwise.
+   */
+  hasPersistedCacheExpired() {
+    const persistedAt = parseInt(
+      localStorage.getItem(this.timePersistedLocalStorageKey) || ''
```
karisal-anders, line 145:
```suggestion
      localStorage.getItem(this.timePersistedLocalStorageKey) ?? ''
```
Unless there's a specific reason for wanting to handle other falsy values beside null and undefined also here? [MSDN](https://developer.mozilla.org/en-US/docs/Web/API/Storage/getItem) says getItem's return value is `A string containing the value of the key. If the key does not exist, null is returned.` so `??` will handle the "not found" case. Also SonarCloud will probably raise a new issue related to https://rules.sonarsource.com/typescript/RSPEC-6606/ i.e. "Nullish coalescing should be preferred".

src/domain/headlessCms/persistor.ts
nikomakela: What else can it catch than Errors, which is a base class?

karisal-anders (approved):
Please take a look at the comments I left. As a possible improvement a test or tests would be good for maintainability and for [KK-1419](https://helsinkisolutionoffice.atlassian.net/browse/KK-1419).

src/domain/headlessCms/persistor.ts
karisal-anders: Probably not much, I think it's theoretically possible for something else like string, number, boolean, undefined, null, object, array or a function to have been raised. It's ok for me if you leave out the else clause here—up to you.

nikomakela: I would say that this evrsion with unknown goes never in else-statement because Error is a base class. Also this is much harder to read and understand. I still think that there is nothing wrong in typing as any in this case.

karisal-anders: If you want to leave it using the `any` type, go ahead. Just thought I'd try to work with Typescript by using the `unknown` type and type narrowing with the `instanceof`. The else clause is probably not that necessary, mostly defensive fallback for theoretical other types.

terovirtanen:
**KUKKUU-UI branch is deployed to platta: https://kukkuu-ui-pr662.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-ui-pr662.dev.hel.ninja :laughing::tada::tada::tada:

karisal-anders:
2 new SonarCloud issues:
image

- "[Member 'logLevel' is never reassigned; mark it as `readonly`.](https://sonarcloud.io/project/issues?pullRequest=662&open=AZUowiqHcmo5mFGtkETH&id=City-of-Helsinki_kukkuu-ui)"
- [Member 'logger' is never reassigned; mark it as `readonly`.](https://sonarcloud.io/project/issues?pullRequest=662&open=AZUowiqHcmo5mFGtkETI&id=City-of-Helsinki_kukkuu-ui)

terovirtanen:
**KUKKUU-UI branch is deployed to platta: https://kukkuu-ui-pr662.dev.hel.ninja** :rocket::rocket::rocket:

